PR #25, opened 2016-05-24
Port to be both Py2 and Py3 compatible.

gramster:
This is valuable as Jupyter is starting to default to Python 3 for new notebooks, and we have had customers trying to use our package in Python 3 environments.

In many files this just involved a couple of extra imports from the Python future package. I had to eliminate circular imports too because the way they can be handled in Python 2 doesn't work in Python 3 and vice-versa. 

The main other change is that strings in Py3 are unicode, so we need to deal with that in a few places.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,5 @@
 .idea
 MANIFEST
 build
-dist
 .coverage
-
+dist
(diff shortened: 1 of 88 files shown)


ojarjur:
LGTM

PR #27, opened 2016-06-06
Monitoring API

supriyagarg:
This initial version of the monitoring client for Datalab allows users
to query timeseries data. It introduces a new module under datalab that
users can import using:

    from datalab.stackdriver import monitoring

Querying timeseries data:

There is a `Query` class that allows users to query timeseries data
for their monitored resources. They can initialize the query by specifying a
metric type and time interval, and refine it by adding filters to it.

The timeseries data is returned as a pandas DataFrame that allows users
to further manipulate the data and visualize it within Datalab.

IPython magics:

There are a couple of IPython magic commands to allow users to list
details of the available metrics and resource types. E.g.:

    %%monitoring list metrics

The base library:

This library is built on top of the gcloud-python monitoring library,
and you can look at its docs here:
https://gcloud-python.readthedocs.io/en/latest/monitoring-usage.html

The Datalab library adds interactive features to it to make it more user
friendly for use in an interactive fashion.

The authors of this code are @rimey and myself, both at Google.

### DIFF
--- a/datalab/kernel/__init__.py
+++ b/datalab/kernel/__init__.py
@@ -31,6 +31,7 @@ import datalab.context as _context
 import datalab.bigquery.commands
 import datalab.context.commands
 import datalab.data.commands
+import datalab.stackdriver.commands
 import datalab.storage.commands
 import datalab.utils.commands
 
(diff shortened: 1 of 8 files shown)


supriyagarg:
Closing this PR, and replacing by an identical PR from a branch (rather than a fork).

PR #35, opened 2016-07-25
Save project to config file so it can persist across kernels and Data…

qimingj:
…lab runs.


ojarjur:
LGTM. Thanks for putting this together.

PR #40, opened 2016-07-31
Allow users to specify whether to use BigQuery legacy SQL or standard SQL

### LINKED ISSUE #39: Add support for BigQuery Standard SQL
BigQuery recently added support for Standard SQL (beta). It would be helpful to have the option to choose between Legacy SQL and Standard SQL.

Links:
https://cloud.google.com/blog/big-data/2016/06/bigquery-111-now-with-standard-sql-iam-and-partitioned-tables
https://cloud.google.com/bigquery/sql-reference/
https://cloud.google.com/bigquery/sql-reference/enabling-standard-sql

parthea:
Closes #39

### DIFF
--- a/datalab/bigquery/_query.py
+++ b/datalab/bigquery/_query.py
@@ -206,22 +206,26 @@ class Query(object):
     """ Get the code for any Javascript UDFs used in the query. """
     return self._code
 
-  def results(self, use_cache=True):
+  def results(self, use_cache=True, dialect='legacy'):
     """Retrieves table of results for the query. May block if the query must be executed first.
 
     Args:
       use_cache: whether to use cached results or not. Ignored if append is specified.
+      dialect : {'legacy', 'standard'}, default 'legacy'
+          'legacy' : Use BigQuery's legacy SQL dialect.
+          'standard' : Use BigQuery's standard SQL (beta), which is
+          compliant with the SQL 2011 standard.
     Returns:
       A QueryResultsTable containing the result set.
     Raises:
       Exception if the query could not be executed or query response was malformed.
     """
     if not use_cache or (self._results is None):
-      self.execute(use_cache=use_cache)
+      self.execute(use_cache=use_cache, dialect=dialect)
     return self._results.results
 
   def extract(self, storage_uris, format='csv', csv_delimiter=',', csv_header=True,
-              compress=False, use_cache=True):
+              compress=False, use_cache=True, dialect='legacy'):
     """Exports the query results to GCS.
 
     Args:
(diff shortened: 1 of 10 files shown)


ojarjur:
Thanks for putting this together. I think this is the right first step, but we should probably keep #39 open until we also address plumbing the new keyword argument into the _bigquery commands.

datalab/bigquery/_query.py
```diff
@@ -332,14 +345,17 @@ def sample(self, count=5, fields=None, sampling=None, use_cache=True):
...
     Raises:
       Exception if the query could not be executed or query response was malformed.
     """
     return Query.sampling_query(self._sql, self._context, count=count, fields=fields,
                                 sampling=sampling, udfs=self._udfs,
-                                data_sources=self._data_sources).results(use_cache=use_cache)
+                                data_sources=self._data_sources).results(use_cache=use_cache,
+                                                                         use_legacy_sql=use_legacy_sql)
 
   def execute_dry_run(self):
```
ojarjur: I think we should also add the same keyword argument ('use_legacy_sql=True') here.

parthea:
@ojarjur Its great to see Travis running here! Note: There are tests failing, but the travis build status is green. I expected the build to fail. I will take a look at the failed tests.

```
Ran 209 tests in 0.348s
FAILED (failures=4)
The command "python ./tests/main.py" exited with 0.
Done. Your build exited with 0.
```

ojarjur:
@parthea You are right about the erroneous success; The test framework was only checking errors, but not failures.

I've gotten that fixed in #50

datalab/bigquery/_query.py
parthea: Done

parthea:
Still need to add support for the new keyword argument in _bigquery

parthea:
Taken from googledatalab/datalab#949 ,

I've confirmed that the following code works with the changes in this pull request:
```
%%sql -d 'standard'
WITH q1 as (
  SELECT count(*)  FROM `myds.logs_*`
)
select * from q1
```

as well as

```
%%sql --module std_sql_sample
WITH q1 as (
  SELECT count(*)  FROM `myds.logs_*`
)
select * from q1
```

followed by 

```
import datalab.bigquery as bq
bq.Query(std_sql_sample).to_dataframe(dialect='standard')
```

My plan for tomorrow is to add unit tests as well as a sample notebook for BigQuery Standard SQL

datalab/bigquery/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2015 Google Inc. All rights reserved.
+# Copyright 2016 Google Inc. All rights reserved.
```
ojarjur: Actually, the year listed here should be the year of the first version, not the latest.

I.E. we never change it for an existing file.

ojarjur:
This LGTM (other than the copyright header changes), so let me know when you feel happy with the unit test changes you wanted to add (I see there are already some in there).

datalab/bigquery/__init__.py
parthea: Done

parthea:
I'm happy with the unit tests in this PR.  I plan to add more unit tests at later stage. (For example, if I encounter an issue that was missed during testing)

I'm also working on a sample notebook. I create a separate PR for that in the googledatalab\notebooks repo.

ojarjur:
LGTM

Thanks, so much for putting in the effort to improve this!

PR #50, opened 2016-08-18
Support Python 3

ojarjur:
We had tried to support Python 3 previously, but we were not testing under Python 3 to ensure that we maintained that support.

This change adds testing for Python 3 and fixes the breakages that have crept in since our previous effort.


qimingj:
LGTM. Thanks!

PR #56, opened 2016-09-20
Update the gcloud library to google-cloud.

supriyagarg:
This is in preparation to adding support for Stackdriver Groups API.


supriyagarg:
@ojarjur: This is a short PR that updates the gcloud python library version after they that had a change in the naming format. I will have some follow up PRs once this is in.

PR #61, opened 2016-09-22
Restore the Sphinx docs

ojarjur:
This change copies over the old docs that used to be in the "datalab" repository, and updates them to reflect the new Python package names.


ojarjur:
@gramster I added you to this pull request just in case you wanted to chime in on it.

PR #64, opened 2016-09-22
Add a QueryMetadata class to the stackdriver monitoring module

supriyagarg:
Add a class that allows users to look at the metadata of a timseries.

* stackdriver/monitoring/_timeseries.py was renamed to stackdriver/monitoring/_query.py. This does not effect the usage
* The method `Query.labels_as_dataframe()` was dropped, and replaced by `Query.metadata()` that returns an object of type `QueryMetadata`.
* To look at the dataframe, users can now run: `query.metadata().as_dataframe()`


tests/stackdriver/monitoring/query_metadata_tests.py
```diff
@@ -0,0 +1,150 @@
...
+
+from __future__ import absolute_import
+import mock
+from oauth2client.client import AccessTokenCredentials
+import unittest
+
+from google.cloud.monitoring import Resource
+from google.cloud.monitoring import Metric
+from google.cloud.monitoring import TimeSeries
+import pandas
```
parthea: This line may not be required. An `unused import` warning appears in my IDE.

supriyagarg: Thanks - done. Also removed a couple of other unused imports.

PR #65, opened 2016-09-22
Stackdriver groups

supriyagarg:
Users can now list groups, via the library, as well as IPython cell magic.

E.g.

    from datalab.stackdriver import monitoring as gcm
    gcm.Groups().as_dataframe()

OR

    % monitoring list groups

### DIFF
--- a/datalab/stackdriver/monitoring/_group.py
+++ b/datalab/stackdriver/monitoring/_group.py
@@ -1,0 +1,88 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License.  You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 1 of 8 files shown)


datalab/stackdriver/monitoring/_group.py
```diff
@@ -0,0 +1,104 @@
...
+        break
+      parent = self._group_dict.get(group.parent_id)
+      parent_display_name = '' if parent is None else parent.display_name
+      data.append([
+          group.id, group.display_name, group.parent_id,
+          parent_display_name, group.is_cluster, group.filter])
+
+    return pandas.DataFrame(data, columns=self._DISPLAY_HEADERS)
+
+  def table(self, pattern='*', max_rows=None):
```
vasbala: Suggest deleting table() and moving pandas.set_option() to the cell magic. The user can visually see the dataframe by using the returned value from as_dataframe() directly in the notebook.

supriyagarg: The pull request #77 removes table method from the resource and metric descriptor classes, and moves the implementation to the cell magic module. The same changes can be made here once that pull request is in.

supriyagarg: @vasbala This PR is uppdated so that the cell magic for listing groups works without the table method.

PR #67, opened 2016-09-23
Support TSV in DataSet.

qimingj:


datalab/ml/_dataset.py
```diff
@@ -373,11 +381,13 @@ def plot(self, names=None, columns=None):
...
         df_correlation = df_correlation[columns]
       for k in df_correlation.columns:
-        if str(df_correlation[k].dtype) == 'object':
-          df_correlation[k] = df_correlation[k].fillna('')
-        elif str(df_correlation[k].dtype) == 'category':
-          df_correlation[k] = df_correlation[k].fillna(df_correlation[k].cat.categories[0])
+        if k == self._target_name:
+          continue
+        elif str(df_correlation[k].dtype) == 'object' or str(df_correlation[k].dtype) == 'category':
+          # pairplot only works with numeric columns
```
ojarjur: This looks like a good change, but I don't understand how it is related to TSV files.

Should this go in the same pull request as the rest of the changes or be its own thing?

qimingj: I can make it a separate change.

qimingj: PTAL.

PR #68, opened 2016-09-23
Support the new BigQuery types in charts

ojarjur:
BigQuery's new Standard SQL support adds additional data [types](https://cloud.google.com/bigquery/sql-reference/data-types) that might occur in the schema field of results.

These new types meant that we would get key errors when looking up the corresponding type to use for charting the data.

Some of those new types correspond to the set of types supported by [Google Charts](https://developers.google.com/chart/interactive/docs/reference#DataTable_addColumn) and some do not.

For the new types that have a correspondence we add that mapping to the code. For any other types, we simply map them to 'string' (to avoid a key error).


datalab/notebook/static/charting.ts
```diff
@@ -440,11 +440,20 @@ module Charting {
...
-        if (data.cols[i].type == 'datetime') {
+        if (data.cols[i].type == 'date' || data.cols[i].type == 'datetime') {
           var rows = data.rows;
           for (var j = 0; j < rows.length; j++) {
             rows[j].c[i].v = new Date(rows[j].c[i].v);
           }
+        } else if (data.cols[i].type == 'timeofday') {
+          var rows = data.rows;
+          for (var j = 0; j < rows.length; j++) {
+            var timeInSeconds = rows[j].c[i].v.split('.')[0];
```
qimingj: Want to make sure "rows[j].c[i].v.split('.')[0]" is the total seconds?

PR #72, opened 2016-09-26
Renamed ml module and command to mlalpha

yebrahim:


nikhilk:
Does setup.py need to change? From brief glance, thats what I think the failed jenkins build is indicating...

PR #73, opened 2016-09-27
Feature/mldev

qimingj:
Update datalab.ml to datalab.mlalpha in references. 
Remove feature preprocessing from Cloud predictor as it is not required.
Optimize pairplot in dataset to remove non numeric columns before plotting.

### DIFF
--- a/datalab/mlalpha/_cloud_predictor.py
+++ b/datalab/mlalpha/_cloud_predictor.py
@@ -32,17 +32,15 @@ _CLOUDML_DISCOVERY_URL = 'https://storage.googleapis.com/cloud-ml/discovery/' \
 class CloudPredictor(object):
   """Preforms cloud predictions on given data."""
 
-  def __init__(self, model_name, version_name, metadata_path=None, label_output=None,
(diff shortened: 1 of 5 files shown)


datalab/mlalpha/_cloud_predictor.py
```diff
@@ -32,29 +32,22 @@
 class CloudPredictor(object):
   """Preforms cloud predictions on given data."""
 
-  def __init__(self, model_name, version_name, metadata_path=None, label_output=None,
+  def __init__(self, model_name, version_name, label_output=None,
```
ojarjur: The 'label_output' keyword parameter is now unused.

I think that either the 'self._label_output = label_output' line was removed on accident, or this should be removed from the constructor (possibly replaced with an '**unused_kwargs' parameter?).

qimingj: We may need to remove it for both local LocalPredictor and CloudPredictor, but have not decided yet. It depends on whether the online prediction returns labels or integers in classification. It is  possible that we will still need it.

I will add a TODO here and probably today will send another change for this once we decided.

PR #80, opened 2016-09-30
Register Datalab specific magic commands as both line and cell magic

parthea:
Partial fix for #71 

The following Datalab magic commands were registered as line magic only:
```
%tensorboard
%monitoring
%storage
%extension
```

If the cell body contains data, line magic commands are automatically converted to cell magic commands.

Since the magic commands listed above are not registered for cell magic, we would see the following error in the Datalab console:
```
{"name":"app","hostname":"b476953e06bc","pid":52,"type":"jupyter","level":30,"msg":"[9000]: ERROR:root:Cell magic `%%tensorboard` not found (But line magic `%tensorboard` exists, did you mean that instead?).\n","time":"2016-09-29T23:56:51.188Z","v":0}
```

This PR should prevent the error from appearing in console.

I think this is only a partial fix for #71 because we still need to notify the user when there is data in the cell body which is not related to the magic command.  For example, a warning should be displayed for the following code snippet since `print(cars)` is unrelated to the `%storage` magic command.
```
%storage read --object gs://cloud-datalab-samples/cars.csv --variable cars
print(cars)
```

### DIFF
--- a/datalab/mlalpha/commands/_tensorboard.py
+++ b/datalab/mlalpha/commands/_tensorboard.py
@@ -20,9 +20,9 @@ import datalab.mlalpha
 import datalab.utils.commands
 
 
-@IPython.core.magic.register_line_magic
-def tensorboard(line):
-  """Implements the tensorboard line magic.
+@IPython.core.magic.register_line_cell_magic
+def tensorboard(line, cell=None):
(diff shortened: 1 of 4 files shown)


qimingj:
Thanks! At some point we will refactor our parsing logic to address #71.

PR #83, opened 2016-10-07
Fix a bug in the passing around of the csv delimiter

ojarjur:
Apparently, the Pandas DataFrame type is special in that it only
supports the keyword argument 'sep' for specifying the delimiter used in
a CSV file, and not the keyword argument 'delimiter'. (Note that the
builtin 'csv' package only supports the keyword argument 'delimiter' and
not 'sep', so we cannot normalize on a single keyword).

Furthermore, somewhere deep in the DataFrame code it checks if the type
of that argument is 'str', which will succeed in Python 3, but not in
Python 2, as we are using Python 3-compatible strings which have a type
of 'unicode', and were overriding the 'str' method with a Python
3-compatible one which returns values with type 'newstr'.

All of this means that in order to pass a single character argument to
the method, we had to distinguish between all of the calls where a
Python 3 compatible str was expected with this one call that requires an
argument with type 'str'. That distinction is being made by changing
the 'from builtins import str' to 'from builtins import str as newstr'.

### DIFF
--- a/datalab/data/_csv.py
+++ b/datalab/data/_csv.py
@@ -14,7 +14,7 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 from builtins import next
-from builtins import str
+from builtins import str as newstr
 from builtins import range
 from builtins import object
 
@@ -68,10 +68,10 @@ class Csv(object):
     return lines
 
   def _is_probably_categorical(self, column):
-    if str(column.dtype) != 'object':
+    if newstr(column.dtype) != 'object':
       # only string types (represented in DataFrame as object) can potentially be categorical
       return False
-    if len(max(column, key=lambda p: len(str(p)))) > 100:
+    if len(max(column, key=lambda p: len(newstr(p)))) > 100:
       return False  # value too long to be a category
     if len(set(column)) > 100:
       return False  # too many unique values to be a category
@@ -96,7 +96,7 @@ class Csv(object):
       return pd.DataFrame(columns=headers)
     columns_size = len(next(csv.reader([lines[0]], delimiter=self._delimiter)))
     if headers is None:
-      headers = ['col' + str(e) for e in range(columns_size)]
+      headers = ['col' + newstr(e) for e in range(columns_size)]
     if len(headers) != columns_size:
       raise Exception('Number of columns in CSV do not match number of headers')
     buf = StringIO()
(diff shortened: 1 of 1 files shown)


datalab/data/_csv.py
```diff
@@ -176,4 +176,4 @@ def sample_to(self, count, skip_header_rows, strategy, target):
         datalab.utils.gcs_copy_file(f.name, target)
     else:
       with open(target, 'w') as f:
-        df.to_csv(f, header=False, index=False, delimiter=self._delimiter)
+        df.to_csv(f, header=False, index=False, sep=str(','))
```
qimingj: Should this be sep=self._delimiter?

ojarjur: Yes, it should. I changed things up when I was debugging why simply casting to `str` wasn't working (because an import was overriding `str`), and I forgot to change it back.

Thanks for catching that.

wafisher:
I'm running into this issue now. Do you mind describing (preferably in a more public forum like the README or the official Datalab docs) how changes like this make their way into the Docker container?

And assuming I don't want to wait for another Datalab version release, what's the best way to get the code from `master` into my locally-run container?

Thanks!

yebrahim:
Take a look at the Dockerfile for datalab's base image [here](https://github.com/googledatalab/datalab/blob/master/containers/base/Dockerfile#L148), the pydatalab repo is cloned (unless you point to a local copy of it, read below) everytime the datalab image is built, this is how these changes get there. I agree this needs to be cleared up in the documentation, perhaps in the install script.

If what you need is to get your own copy of this repo (pydatalab) into your locally-built datalab image then it's easier to do it while building the datalab image. To do this you can pass the path to your local pydatalab to the build script at containers/datalab/build.sh, for example:
`./build.sh ../../pydatalab/` < note the slash in the end

The install instructions on this repo are meant for building Jupyter nbextensions, which can then be used from any Jupyter instance.

wafisher:
Thank you. What version of Python should I be running Datalab and this extension with or does it not matter?

parthea:
> Also, what I'm going to try to do is install this locally into site packages and load it in the kernel — is that a bad idea?

I often use this approach because I find it faster than rebuilding the image, although it may not be supported officially.  

Here are the steps:

- Clone pydatalab using `git clone https://github.com/googledatalab/pydatalab.git`
- `cd` to the folder and run either `./install-no-virtualenv.sh` or `install-virtualenv.sh` depending on your environment. Make sure the install completes successfully without errors.
- Replace the copy of pydatalab in the container with your local copy, by using : `docker run -it -p "127.0.0.1:8081:8080" -v "${HOME}:/content"    -e "PROJECT_ID=<your-project>" -v "<path-to-pydatalab-clone>/datalab/:/usr/local/lib/python2.7/dist-packages/datalab/"  gcr.io/cloud-datalab/datalab:local`

I hope this helps.

PR #90, opened 2016-10-27
Remove no longer needed mathjax extension

yebrahim:
MathJax is not added by default for all notebooks, it's not needed as an extension. We'll keep it for now for backward compatibility.


ojarjur:
It looks like there are also references to this file that need to be removed before the tests will pass.

PR #94, opened 2016-11-17
Add ability to set the default BigQuery SQL dialect

### LINKED ISSUE #93: Add an option to set the default dialect to Standard SQL
#40 added an option to specify whether to use the standard SQL dialect, but this is done on a per-query basis. On notebooks with lots of queries specifying dialect=standard each time is cumbersome and error prone, it'd be nice to have an option to change the default on a per-notebook level.

parthea:
Closes #93

By default, Datalab will set the BigQuery SQL dialect to legacy for all queries.

Use `set_datalab_bq_dialect('standard')` to set the default BigQuery SQL dialect to Standard SQL for the notebook, and `datalab_bq_dialect()` to read the BigQuery SQL dialect specified for the notebook.

Note: If the dialect parameter is provided on a per query basis, then the dialect parameter for that query will be honored (i.e. we will ignore the notebook default).

For example,
If the SQL dialect for the notebook is set to legacy, the following query will still execute successfully since the dialect is specified on a per query basis.
```
%%sql -d standard
WITH UniqueNames2013 AS
(SELECT DISTINCT name
  FROM `bigquery-public-data.usa_names.usa_1910_2013`
  WHERE Year = 2013)
SELECT COUNT(*) FROM UniqueNames2013
```

### DIFF
--- a/datalab/bigquery/__init__.py
+++ b/datalab/bigquery/__init__.py
@@ -15,6 +15,7 @@ from __future__ import absolute_import
 
 from ._csv_options import CSVOptions
 from ._dataset import Dataset, Datasets
+from ._dialect import Dialect
 from ._federated_table import FederatedTable
 from ._job import Job
 from ._query import Query
--- a/datalab/bigquery/_api.py
+++ b/datalab/bigquery/_api.py
@@ -17,6 +17,7 @@ from past.builtins import basestring
 from builtins import object
 
 import datalab.utils
+import datalab.bigquery
 
 
 class Api(object):
@@ -124,7 +125,7 @@ class Api(object):
 
   def jobs_insert_query(self, sql, code=None, imports=None, table_name=None, append=False,
                         overwrite=False, dry_run=False, use_cache=True, batch=True,
-                        allow_large_results=False, table_definitions=None, dialect='legacy',
+                        allow_large_results=False, table_definitions=None, dialect=None,
                         billing_tier=None):
     """Issues a request to insert a query job.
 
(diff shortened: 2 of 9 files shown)


qimingj:
Thanks Anthonios!

We plan to make standard SQL the default so maybe we don't need the global dialect setting. What do you think?

qimingj:
Assigning to Chris.

parthea:
My initial thought is that a global setting could be useful regardless of whether legacy or standard is the default. 

My feeling is that with this change there will be less impact to existing users after changing the default dialect to Standard SQL. They would simply have `set_datalab_bq_dialect('legacy')` in a cell in their notebook, as opposed to setting the dialect to `'legacy'` on a per-query basis for all queries.

parthea:
> We plan to make standard SQL the default 

Can I go ahead and make that change in this PR?

qimingj:
My only concern is that we are adding it to context class, and in the future we will probably need to remove it so it will become a backward compatibility issue.

Can we move the setting to a global variable such as datalab.bigquery._dialect, so you could still use it but it will be easier to remove in the future?

parthea:
> Can we move the setting to a global variable such as datalab.bigquery._dialect, so you could still use it but it will be easier to remove in the future?

Yes. I'll make this change. Let me know if it is ok to change the default dialect to Standard SQL here. If we don't want to make the change now, I expect it will only be a 1 line change to make standard SQL default in another PR.

qimingj:
Let's not change the default dialect since it is another breaking change.

parthea:
The changes in this PR are ready for review.

qimingj:
Thanks Anthonios!

PR #100, opened 2016-12-01
Remove unused imports

parthea:
This PR removes unused imports as reported by [flake8](https://pypi.python.org/pypi/flake8) with code [F401: module imported but unused](http://flake8.pycqa.org/en/latest/user/error-codes.html?highlight=f401)

When I ran `flake8`, I excluded `__init__.py` files by adding `# flake8: noqa` to the top of those files. I didn't include that change in this PR, although it can be added in another PR that enables flake8 on Travis (if decided). Here is the output from `flake8 --select=F401` after ignoring `__init__.py`

```
tony@tonypc:~/pydatalab-parthea/datalab$ flake8 --select=F401
./storage/_item.py:264:7: F841 local variable '_' is assigned to but never used
./storage/_bucket.py:229:7: F841 local variable '_' is assigned to but never used
./data/_csv.py:164:35: F821 undefined name 'xrange'
```

### DIFF
--- a/datalab/bigquery/_table.py
+++ b/datalab/bigquery/_table.py
@@ -22,7 +22,6 @@ import calendar
 import codecs
 import csv
 import datetime
-import math
 import pandas
 import time
 import traceback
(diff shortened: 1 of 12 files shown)


parthea:
:+1: Lightning fast!

PR #101, opened 2016-12-05
Allow users to save charts as a PNG file or in a python variable

parthea:
Closes googledatalab/datalab#783

When using the `%%chart` magic command, allow users to export the chart image as either PNG or in a python variable as a base64 encoded string.

To save the chart as a PNG file, use the following sample code:

```
data = [
        {'date': '2016/01/01', 'count': 10}, 
        {'date': '2016/01/02', 'count': 35}, 
        {'date': '2016/01/03', 'count': 25} 
       ]
%chart pie -d data --output my_cool_chart.png
```

Alternatively, specify the name of a python variable where the chart data will be exported in base64 encoded string format.

```
data = [
        {'date': '2016/01/01', 'count': 10}, 
        {'date': '2016/01/02', 'count': 35}, 
        {'date': '2016/01/03', 'count': 25} 
       ]
%chart pie -d data --output my_cool_chart
```

Followed by:
```
from IPython.display import Image
Image(my_cool_chart) 
```

screenshot from 2016-12-04 23-01-22

### DIFF
--- a/datalab/bigquery/commands/_bigquery.py
+++ b/datalab/bigquery/commands/_bigquery.py
@@ -946,7 +946,8 @@ def _table_viewer(table, rows_per_page=25, fields=None):
             }},
             {{source_index: {source_index}, fields: '{fields}'}},
             0,
-            {total_rows});
+            {total_rows},
+            '');
         }}
       );
     </script>
--- a/datalab/notebook/static/charting.ts
+++ b/datalab/notebook/static/charting.ts
@@ -409,6 +409,7 @@ module Charting {
     hasIPython:boolean;
     cellElement:HTMLElement;
     totalRows:number;
+    outputLocation:string;
 
     constructor(protected driver:ChartLibraryDriver,
                 protected dom:Element,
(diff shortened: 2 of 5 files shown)


qimingj:
Anthonios, thanks for another contribution!

With this change, is it possible that opening a notebook (with the cell executed and output available) will invoke an image saving operation, even without user executing the cell again? The saving operation is triggered from client side script as an kernel call, so when the kernel is connected it is possible that refreshing the page may trigger it unexpectedly.

parthea:
> With this change, is it possible that opening a notebook (with the cell executed and output available) will invoke an image saving operation, even without user executing the cell again?

Yes, I agree this is an issue. I'm going to spend more time on this PR this evening to try and find a solution. If you have any suggestions on potential solutions for this issue, please don't hesitate to share any ideas! It would be really helpful. 

Also, after testing the code again this morning, I found that the 'export to Python variable' functionality only works if you run cells individually. I found that the export to variable function is not needed. I'm going to change this PR to only export the chart as a PNG file .

qimingj:
Is it possible to generate the chart image at kernel so we don't have to send the image bytes from client, which also fixes the issue?

parthea:
Thanks for the suggestion! I'm going to investigate this option. This PR should re-open again when I push a new version.

Issue #106, opened 2016-12-08
Typescript Error while building pydatalab: Duplicate identifier '_this'


parthea:
I notice the following error appears when I try to build pydatalab on Ubuntu 16.10. 
```
tony@tonypc:~/pydatalab-parthea$ ./install-no-virtualenv.sh 
datalab/notebook/static/charting.ts(810,11): error TS2399: Duplicate identifier '_this'. Compiler uses variable declaration '_this' to capture 'this' reference.
datalab/notebook/static/charting.ts(812,9): error TS2400: Expression resolves to variable declaration '_this' that compiler uses to capture 'this' reference.
```

This error is also causing the Datalab build script `$REPO_DIR/containers/datalab/build.sh` to fail.
```
Step 14 : RUN ipython profile create default &&     jupyter notebook --generate-config &&     if [ -d /datalab/lib/pydatalab/.git ]; then         echo "use local lib";       else         git clone https://github.com/googledatalab/pydatalab.git /datalab/lib/pydatalab;       fi &&     cd /datalab/lib/pydatalab &&     /tools/node/bin/npm install -g typescript &&     tsc --module amd --noImplicitAny --outdir datalab/notebook/static datalab/notebook/static/*.ts &&     /tools/node/bin/npm uninstall -g typescript &&     pip install --no-cache-dir . &&     jupyter nbextension install --py datalab.notebook &&     jupyter nbextension enable --py widgetsnbextension &&     rm datalab/notebook/static/*.js &&     mkdir -p /datalab/nbconvert &&     cp -R /usr/local/share/jupyter/nbextensions/gcpdatalab/* /datalab/nbconvert &&     ln -s /usr/local/lib/python2.7/dist-packages/notebook/static/custom/custom.css /datalab/nbconvert/custom.css &&     mkdir -p /usr/local/lib/python2.7/dist-packages/notebook/static/components/codemirror/mode/text/sql/text &&     ln -s /usr/local/share/jupyter/nbextensions/gcpdatalab/codemirror/mode/sql.js /usr/local/lib/python2.7/dist-packages/notebook/static/components/codemirror/mode/text/sql/text/sql.js &&     cd /
 ---> Running in 6ade2d2451af
[ProfileCreate] Generating default config file: u'/root/.ipython/profile_default/ipython_config.py'
[ProfileCreate] Generating default config file: u'/root/.ipython/profile_default/ipython_kernel_config.py'
[...]

PR #107, opened 2016-12-08
Use -e in all shell scripts to fail on error

parthea:
Use -e in all shell scripts to fail on error.  It was difficult to spot the error in #106 while building pydatalab. I actually noticed the error while building Datalab because the Datalab build scripts already have `-e`.


parthea:
Travis will fail for this PR until #106 is merged

yebrahim:
Thanks for taking care of this.

PR #113, opened 2016-12-29
Generate reST documentation for magic commands

yebrahim:
This PR adds docs for any added magics automatically by searching through the source files for lines with `register_line_cell_magic`, capturing the names for those magics, and calling them inside an `ipython` kernel with the `-h` argument, then storing that output into a generated `datalab.magics.rst` file.

I also fixed the `README` with more complete instructions, and fixed the `pip install` dependency order, which was causing installing `scikit-learn` to fail because of a versioning issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ MANIFEST
 build
 .coverage
 dist
+datalab.magics.rst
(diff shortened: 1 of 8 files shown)


ojarjur:
This looks really good, thanks for getting that to work!

PR #122, opened 2017-01-11
Remove some unused dependencies + Add some others

eyadsibai:
- Remove scikit-learn and scipy from pydatalab package. They are not used anywhere in this repo, but they might be used in the notebooks so either remove them or not having a specific version.

- Add plotly, seaborn, psutil and google api client packages as they are being used in this repo.


setup.py
```diff
@@ -71,16 +71,18 @@
     'future==0.15.2',
     'futures==3.0.5',
     'google-cloud==0.19.0',
+    'google-api-python-client==1.6.0',
```
ojarjur: Thanks so much for catching this!

We absolutely should include these dependencies. The versions we've tested, however are still a bit behind the ones you've selected. Would you be willing to change to match the versions we currently use in our [Dockerfile](https://github.com/googledatalab/datalab/blob/master/containers/base/Dockerfile)?

Those versions would be:

    google-api-python-client==1.5.1
    seaborn==0.7.0
    plotly==1.12.5
    psutil==4.3.0

setup.py
```diff
@@ -71,16 +71,18 @@
...
     'httplib2==0.9.2',
     'oauth2client==2.2.0',
+    'psutil==5.0.1',
     'pandas>=0.17.1',
     'pandas-profiling>=1.0.0a2',
     'python-dateutil==2.5.0',
     'pytz>=2015.4',
     'pyyaml==3.11',
     'requests==2.9.1',
-    'scipy==0.18.0',
```
ojarjur: These two are not directly used by our code, but I think that we've included them because we want any of our users to be able to use them out of the box.

@chmeyers and @qimingj might know better about that, though.

yebrahim: In that case I think it should be part of Datalab (and it is) but not pydatalab right?

setup.py
eyadsibai: Sure

setup.py
qimingj: Yes. They are included for user's convenience. You don't want users to install them every time they start Datalab. 
Maybe we don't need to put the same list twice in docker and setup.py? Installation of pydatalab will install these anyway. That way we leave only the packages that are not listed in setup.py.

yebrahim: This is all related to datalab though, pydatalab shouldn't be related to that convenience.

PR #123, opened 2017-01-11
Cleanup

eyadsibai:
- Remove unnecessary semicolons
- Remove unused imports
- Remove unnecessary defined variable that it defined again in couple of lines.


datalab/storage/_item.py
```diff
@@ -196,7 +196,6 @@ def read_lines(self, max_lines=None):
 
     max_to_read = self.metadata.size
     bytes_to_read = min(100 * max_lines, self.metadata.size)
-    lines = []
```
yebrahim: Why remove this?

datalab/mlalpha/commands/_mlalpha.py
```diff
@@ -255,7 +254,7 @@ def _train(args, cell):
         urllib.urlencode(log_url_query_strings)
     html += '<p>Click <a href="%s" target="_blank">here</a> to view cloud log. <br/>' % log_url
     html += 'Start TensorBoard by running "%tensorboard start --logdir=&lt;YourLogDir&gt;".</p>'
-    return IPython.core.display.HTML(html);
```
yebrahim: We should probably add flake8 to pydatalab, it would've caught this.

datalab/storage/_item.py
eyadsibai: not needed ... it's defined again in line 203

PR #124, opened 2017-01-11
Add cloud ml dependency

eyadsibai:
- cloud ml is being used in this package. This will make pydatalab works fine and pip installable.

- without this package if you try to %load_ext datalab.kernel, you will get an error message saying that google.cloud.ml does not exist. So you won't be able to use cell magic things.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -83,6 +83,9 @@ for accessing Google's Cloud Platform services such as Google BigQuery.
     'scikit-learn==0.17.1',
     'ipykernel==4.4.1',
   ],
+  dependency_links=[
+     "https://storage.googleapis.com/cloud-ml/sdk/cloudml.latest.tar.gz",
+    ],
   package_data={
     'datalab.notebook': [
         'static/bigquery.css',


yebrahim:
This is great, I think we should take out Python3 from travis anyway since we don't officially support it yet. @qimingj what do you think?

eyadsibai:
It seems it is working with python3. It could be that it supports python3.4 but not python3.5. I did not manage to install it on python3.5

qimingj:
I would love to see pydatalab python3 compatible but one big problem now is Cloud DataFlow (apache beam) is not. Unfortunately Datalab and CloudML SDK have to depend on them, so I don't see another choice.

setup.py
```diff
@@ -83,6 +83,9 @@
     'scikit-learn==0.17.1',
     'ipykernel==4.4.1',
   ],
+  dependency_links=[
+     "https://storage.googleapis.com/cloud-ml/sdk/cloudml.latest.tar.gz",
```
yebrahim: Don't we need 'cloudml' under install_requires?

eyadsibai: you can't add links to install_requires. Cloud ml is not published through pypi.

yebrahim: I mean in addition to the dependency link, we also need 'cloudml' as a requirement.

eyadsibai: True, here is the problem is that there is a package in pypi called cloudml

eyadsibai:
I will close this pull request since this would not work because cloudml package has a conflict name with another package in pypi. I will check if there is a solution

PR #125, opened 2017-01-12
Make the library pip-installable

yebrahim:
This PR adds tensorflow and cloudml in setup.py to make the lib pip-installable. I had to install them explicitly using pip from inside the `setup.py` script, even though it's not a clean way to do it, it gets around the two issues we have at the moment with these two packags:

- Pypi has Tensorflow version 0.12, while we need 0.11 for the current version of pydatalab. According to the [Cloud ML docs](https://cloud.google.com/ml/docs/how-tos/getting-set-up#installing_the_google_api_client_for_python), that version exists as a pip package for three supported platforms.
- Cloud ML SDK exists as a pip package, but also not on Pypi, and while we could add it as a dependency link, there exists another package on Pypi called `cloudml`, and pip ends up installing that instead (see https://github.com/googledatalab/pydatalab/pull/124). I cannot find a way to force pip to install the package from the link I included.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,13 @@
 language: python
 python:
   - "2.7"
-  - "3.4"
 
 before_install:
+  - sudo apt-get install -y python-setuptools
   - npm install -g typescript
   - tsc --module amd --noImplicitAny --outdir datalab/notebook/static datalab/notebook/static/*.ts
   - pip install -U pip
+  - pip install -U setuptools
   - pip install .
 
 script: python ./tests/main.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
-  - "2.7"
+  - 2.7
+  - 3.5
 
 before_install:
   - sudo apt-get install -y python-setuptools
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,12 +1,13 @@
 language: python
 python:
   - "2.7"
-  - "3.4"
```
ojarjur: Instead of removing Python 3 support, can we conditionally install the CloudML and Tensorflow libraries only if the version of Python is Python 2?

I.E. have everything work on Python 2, but if you want Python 3, you can still use the pydatalab library with storage or bigquery.

yebrahim: For some reason, this change makes two of the tests fail with Python 3. When running the tests locally they fail all the time actually, so there seems to be a bug with the travis environment, these tests should never pass in the first place, I opened https://github.com/googledatalab/pydatalab/issues/127 to track this. Meanwhile, I think we should go ahead with this change, disabling Python 3 until we fix that test, since we're not officially supporting Python 3 yet anyway.

yebrahim: https://github.com/googledatalab/pydatalab/pull/128 fixes the tests for Python 3, please take a look at that first.

Issue #127, opened 2017-01-14
query_metadata_tests failing on py3.5


yebrahim:
Update your interpreter locally to 3.5 or use travis 3.5 and these tests fail with the following error:
```
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.5.2/lib/python3.5/site-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/home/travis/build/yebrahim/pydatalab/tests/stackdriver/monitoring/query_metadata_tests.py", line 70, in test_as_dataframe
    dataframe = query_metadata.as_dataframe()
  File "/home/travis/virtualenv/python3.5.2/lib/python3.5/site-packages/datalab/stackdriver/monitoring/_query_metadata.py", line 68, in as_dataframe
    for ts in self._timeseries_list[:max_rows]]
  File "/home/travis/virtualenv/python3.5.2/lib/python3.5/site-packages/datalab/stackdriver/monitoring/_query_metadata.py", line 68, in <listcomp>
    for ts in self._timeseries_list[:max_rows]]
AttributeError: 'Resource' object has no attribute '__dict__'
```

PR #128, opened 2017-01-15
Fix query_metadata tests

yebrahim:
Tests are failing in Python3 because of using `__dict__`. It's not obvious to me why this isn't available for Python3 from the cloudml sdk, but `_asdict()` works for both.

Fixes https://github.com/googledatalab/pydatalab/issues/127.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 python:
   - "2.7"
-  - "3.4"
+  - "3.5"
 
 before_install:
   - npm install -g typescript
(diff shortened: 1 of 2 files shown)


yebrahim:
I found the answer after some digging, this indeed is a difference between Python2 and Python3, a nuisant one at that.

It turns out the type Resource in the code is actually a named tuple. In Python2, `.__dict__` is added as convenient property wrapper of `._asdict()`, but this isn't done in Python3:

https://docs.python.org/2/library/collections.html#collections.somenamedtuple._asdict (find `__dict__` here) https://docs.python.org/3/library/collections.html#collections.somenamedtuple._asdict

So it looks like `._asdict()` is in fact the source of truth, and should be used for portable 2to3 code.

PR #129, opened 2017-01-18
Remove old feature-slicing pipeline implementation (is replaced by BigQuery)  Add Confusion matrix magic.

qimingj:

### DIFF
--- a/datalab/mlalpha/_analysis.py
+++ b/datalab/mlalpha/_analysis.py
@@ -10,206 +10,39 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 
-"""Implements Cloud ML Eval Results Analysis"""
-
-import apache_beam as beam
-from collections import namedtuple
-
-"""Prepresents an eval results CSV file. For example, the content is like:
-     107,Iris-versicolor,1.64827824278e-07,0.999999880791,6.27104979056e-10
-     100,Iris-versicolor,3.5338824091e-05,0.99996471405,1.32811195375e-09
-     ...
-"""
-CsvEvalResults = namedtuple('CsvEvalResults', 'source, key_index predicted_index score_index_start num_scores')
-
-"""Prepresents an eval source CSV file. For example, the content is like:
-     107,Iris-virginica,4.9,2.5,4.5,1.7
-     100,Iris-versicolor,5.7,2.8,4.1,1.3
-     ...
-   The metadata is generated in the preprocessing pipeline. It is used to describe the CSV file,
-   including schema, headers, etc.
-"""
-CsvEvalSource = namedtuple('CsvEvalSource', 'source metadata')
+"""Implements Cloud ML Analysis Helpers"""
 
 
-class EvalResultsCsvCoder(beam.coders.Coder):
-  """A coder to read from Eval results CSV file. Note encode() is only needed in cloud run.
-  """
-  def __init__(self, eval_results):
-    self._eval_results = eval_results
+import google.cloud.ml as ml
+import numpy as np
+import pandas as pd
+import yaml
 
-  def decode(self, csv_line):
-    import csv
-    source_elem = next(csv.reader([csv_line]))
-    key = source_elem[self._eval_results.key_index]
-    element = {
-        'predicted': source_elem[self._eval_results.predicted_index],
-        'scores': source_elem[self._eval_results.score_index_start: \
-            self._eval_results.score_index_start+self._eval_results.num_scores]
-    }
-    return (key, element)
-
-  def encode(self, element):
-    return str(element)
+import datalab.bigquery as bq
 
 
-class AccuracyFn(beam.CombineFn):
-  """A transform to compute accuracy for feature slices.
-  """
-  def __init__(self, target_column_name):
-    self._target_column_name = target_column_name
-
-  def create_accumulator(self):
-    return (0.0, 0)
-
-  def add_input(self, (sum, count), input):
-    new_sum = sum
-    if (input['predicted'] == input[self._target_column_name]):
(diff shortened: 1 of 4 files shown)


datalab/mlalpha/_analysis.py
```diff
@@ -10,206 +10,39 @@
...
-      pipeline_option = self.default_pipeline_options(output_dir)
-    opts = beam.pipeline.PipelineOptions(flags=[], **pipeline_option)
-    p = beam.Pipeline('DataflowPipelineRunner', options=opts)
-    self._pipeline_def(p, eval_source, eval_results, features_to_slice, metrics, output_file)
-    p.run()
-
+      Loaded pandas dataframe.
+  """
+  with ml.util._file.open_local_or_gcs(schema_path, mode='r') as f:
+    schema = yaml.load(f)
```
brandondutra: yaml.save_load(f)?

datalab/mlalpha/_analysis.py
```diff
@@ -10,206 +10,39 @@
...
-    """Get default DataFlow options. Users can customize it further on top of it and then
-       send the option to run_cloud().
+def csv_to_dataframe(csv_path, schema_path):
+  """Given a CSV file together with its BigQuery schema file in yaml, load
+     content into a dataframe.
 
     Args:
-      output_dir: A GCS path which will be used as base path for tmp and staging dir.
+      csv_path: Input CSV path. Can be local or GCS.
+      schema_path: Input schema path. Can be local or GCS.
```
brandondutra: can you give an example of the schema file in the comment? 

After loading the schema file, the order of the column names in the python object is preserved, right?

datalab/mlalpha/commands/_ml.py
```diff
@@ -158,3 +181,57 @@ def _predict(args, cell):
...
+  plt.colorbar()
+  tick_marks = np.arange(len(labels))
+  plt.xticks(tick_marks, labels, rotation=45)
+  plt.yticks(tick_marks, labels)
+  plt.tight_layout()
+  plt.ylabel('True label')
+  plt.xlabel('Predicted label')
+
+
+def _confusion_matrix(args, cell):
```
brandondutra: this function could call two helpers to make it more readable like _cm_from_csv and _cm_from_sql

qimingj:
We are not moving away from big data. However, we are moving away from DataFlow based eval analysis. The idea is: use CloudML's batch prediction service to generate eval results (possibly sharded CSV files), load the CSV content into BigQuery table, and use BigQuery to perform analysis. 

This change includes local analysis of CSV, mostly because the cloud_batch_prediction() in inception package is not done yet so we only have local batch_prediction(). My plan is that the output of cloud_batch_prediction() will be saved in BigQuery table only (hence no csv), and then the following visualization components based on BigQuery can be plugged in:

%%ml confusion_matrix [--bqtable | --sql]
%%ml feature_slice_view [--sql]

If needed, we can support big CSV files too but the implementation might be through BigQuery federated table (DataFlow is inefficient and not interactive here).

datalab/mlalpha/_analysis.py
qimingj: The schema would be:
- name: target
   type: STRING
- name: predicted
   type: STRING
...
It's a list of dict, so it is ordered even after loaded in memory.

datalab/mlalpha/commands/_ml.py
```diff
@@ -158,3 +181,57 @@ def _predict(args, cell):
...
+  if args['csv'] is not None:
```
brandondutra: Ah, ok. So for now the focus is on building a confusion_matrix from bigquery, and package authors are responsible for getting the results into bigquery. CVS is offered for smaller datasets. Please add a todo saying something like "add option to load csv data into bigquery". Or somehow document that the csv path is not recommended.

datalab/mlalpha/commands/_ml.py
```diff
@@ -158,3 +181,57 @@ def _predict(args, cell):
...
+    _plot_confusion_matrix(cm, labels)
+  elif args['bqtable'] is not None or args['sql'] is not None:
+    if args['sql'] is not None:
+      item = datalab.utils.commands.get_notebook_item(args['sql'])
+      query, _ = datalab.data.SqlModule.get_sql_statement_with_environment(item, {})
+    else:
+      query = ('select target, predicted, count(*) as count from %s group by target, predicted'
+               % args['bqtable'])
+    dfbq = datalab.bigquery.Query(query).results().to_dataframe()
+    labels = sorted(set(dfbq['target']))
```
brandondutra: This assumes the labels for the target and predicted column are all the same. This might not be true; for example, what if I run prediction on a test set that is missing a class label. 

I think you want something like "labels = sorted(set(dfbq['target'])+set(dfbq['predicted']))" This still could miss labels if they never show up in the evaluation set or prediction set, but I guess this is ok (but still add a comment)

datalab/mlalpha/commands/_ml.py
qimingj: Added TODO. Will address that TODO once service supports "offline batch prediction".

datalab/mlalpha/commands/_ml.py
qimingj: Good idea. Done.

datalab/mlalpha/_analysis.py
qimingj: Done.

datalab/mlalpha/commands/_ml.py
qimingj: Good catch. Done.

qimingj:
Brandon, thanks for all feedback. PTAL.

PR #132, opened 2017-01-18
add structure data preprocessing and training 

brandondutra:
This solution is not finished, just splitting the pull request now to start the review process. 

In this pull request, the preprocessing code works via gcloud and the datalab interface for both local and cloud runs. Only local training is currently connected in datalab, and it requires TF 0.12 for now. 

The structured data solution solves classification and regression problems using tf.lear DNN and linear models. A new file, for now called transforms config file, is used to define the transforms. This one transforms file is used in preprocessing and training. I'll give an example a following message.

### DIFF
--- a/solutionbox/structured_data/datalab_solutions/structured_data/_package.py
+++ b/solutionbox/structured_data/datalab_solutions/structured_data/_package.py
@@ -23,12 +23,22 @@
    Datalab will look for functions with the above names.
 """
 
+import datetime
 import logging
 import os
-import urllib
+import shutil
 import subprocess
 import sys
-import datetime
+import tempfile
+import urllib
+
+import tensorflow as tf
+import yaml
+
+import google.cloud.ml as ml
+
+_TF_GS_URL = 'gs://cloud-datalab/deploy/tf/tensorflow-0.12.0rc0-cp27-none-linux_x86_64.whl'
+
 
 def _percent_flags(train_percent=None, eval_percent=None, test_percent=None):
   """Convert train/eval/test percents into command line flags.
@@ -81,17 +91,17 @@ def _run_cmd(cmd):
 
 def local_preprocess(input_file_path, output_dir, transforms_config_file,
                      train_percent=None, eval_percent=None, test_percent=None):
-  """Preprocess data locally with Beam. 
+  """Preprocess data locally with Beam.
 
-  Produce output that can be used by training efficiently. Will also split 
+  Produce output that can be used by training efficiently. Will also split
   data into three sets (training, eval, and test). {train, eval, test}_percent
-  should be nonnegative integers that sum to 100. 
+  should be nonnegative integers that sum to 100.
 
   Args:
-    input_file_path: String. File pattern what will expand into a list of csv 
+    input_file_path: String. File pattern what will expand into a list of csv
         files. Preprocessing will automatically slip the data into three sets
-        for training, evaluation, and testing. Can be local or GCS path. 
-    output_dir: The output directory to use; can be local or GCS path. 
+        for training, evaluation, and testing. Can be local or GCS path.
+    output_dir: The output directory to use; can be local or GCS path.
     transforms_config_file: File path to the config file.
     train_percent: Int in range [0, 100].
     eval_percent: Int in range [0, 100].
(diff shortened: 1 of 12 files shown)


brandondutra:
**Example of using cloud preprocessing with datalab magics**

I have two csv data files (gs://cloud-ml-dev_bdt/data*.csv) that contains a regression problem with 3 numeric input columns and 3 categorical input columns. 

The transforms_config_file is located here: gs://cloud-ml-dev_bdt/transforms_config_dnn.json

Note that the config file lists the columns in the csv file, the key columns, the target column, problem type [classification/regression], the model type [dnn/linear], and the transforms [scale, embedding, sparse, etc].

To try cloud preprocessing, 
1) build the package (python setup.py sdist)
2)  run the magic
%ml preprocess --package /content/pydatalab/solutionbox/structured_data/dist/structured_data-0.1.tar.gz --cloud
input_file_path: gs://cloud-ml-dev_bdt/data*.csv
output_dir: gs://BUCKET/preprocessed_output
transforms_config_file: gs://cloud-ml-dev_bdt/transforms_config_dnn.json

Local training only works if tf 0.12 is installed for now.

brandondutra:
**roadmap/todo for future pull requests**
1) get training within datalab finished
2) add prediction code that works off gcloud
3) add the prediction datalab interface
4) revisit the config file: consider splitting it up into a schema/transform file. More thought needs to be made on this.
4) revisit using tf.learn/tf.layers: do not use contrib code

solutionbox/structured_data/datalab_solutions/structured_data/trainer/task.py
```diff
@@ -0,0 +1,357 @@
...
+
+    # Get the model to train.
+    estimator = get_estimator(output_dir, feature_columns, metadata, 
+                              transform_config, args)
+
+    input_placeholder_for_prediction = get_placeholder_input_fn(metadata, 
+        transform_config)
+
+    # Save the finished model to output_dir/model
+    export_monitor = util.ExportLastModelMonitor(
```
qimingj: Does it export the metadata as well? Prediction relies on the metadata in the same dir as the model.

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
```diff
@@ -0,0 +1,236 @@
...
+  """
+  
+  #TODO(brandondutra): fix to TF 0.12
+  #TODO(brandondutra): allow other flags to be set.
+
+  train_filename = os.path.join(preprocessed_dir, 'features_train*')
+  eval_filename = os.path.join(preprocessed_dir, 'features_eval*')
+  metadata_filename = os.path.join(preprocessed_dir, 'metadata.json')
+  this_folder = os.path.dirname(os.path.abspath(__file__))
+  cmd = ['cd %s &&' % this_folder, 'gcloud alpha ml local train',
```
brandondutra: Side question: could reset session also kill any running child process? That sounds like something useful others would need too.

Are you recommending something like this:
local_preprocess: call code directly 
cloud_preprocess: call code directly 
local_train: call code directly 
cloud_train: use datalab/mlalpha/_cloud_runner.py
local_predict: call code directly
cloud_predict: use datalab/mlalpha/_cloud_predictor.py
local_batch_predict: call code directly
cloud_batch_predict: Does datalab support this? Is the only option gcloud?

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
```diff
@@ -0,0 +1,236 @@
...
+         '--metadata_path=%s' % metadata_filename,
+         '--output_path=%s' % output_dir,
+         '--transforms_config_file=%s' % transforms_config_file,
+         '--max_steps=%s' % str(max_steps)]
+  if layer_sizes:
+    cmd += ['--layer_sizes %s' % layer_sizes]
+
+  print('Local training, running command: %s' % ' '.join(cmd))
+  _run_cmd(' '.join(cmd))
+
```
brandondutra: done

solutionbox/structured_data/datalab_solutions/structured_data/trainer/task.py
brandondutra: Yes. "additional_assets=[args.metadata_path]" puts a copy of the metadata.json file in the final model dir.

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
brandondutra: Interesting note: in your %%ml magics doc, in the section called 'Model/Job Management' Nikhil made a comment to use gcloud for the job management stuff. If gcloud is going to be a first class concept for datalab, then maybe there is some value in showing the user the gcloud command for training/prediction. But this still does not address your hanging question.

PR #139, opened 2017-01-25
structured data is done

brandondutra:
--added prediction
--split config files into a schema and transforms file


brandondutra:
example notebook using the ml magics:

gs://cloud-ml-dev_bdt/magical\ structured\ data\ package.ipynb

brandondutra:
example schema file:
{
  "column_names": ["key", "t", "num1", "num2", "num3", "str1", "str2", "str3"],
  "key_column": "key",
  "target_column": "t",
  "numerical_columns": ["num1", "num2", "num3"],
  "defaults": {"num1": 32},
  "categorical_columns": ["t", "str1", "str2", "str3"]
 }

example transforms file for use with DNN models: (not the use of default transforms as some features are not listed here)
{
  "num1": {"transform": "scale"},
  "num2": {"transform": "max_abs_scale",
           "value": 4},
  "str1": {"transform": "one_hot"},
  "str2": {"transform": "embedding",
           "dimension": 3}
 }

qimingj:
Thanks. The notebooks and config examples are very helpful.

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
```diff
@@ -220,7 +210,9 @@ def local_train(preprocessed_dir, transforms_config_file, output_dir,
          '--eval_data_paths=%s' % eval_filename,
          '--metadata_path=%s' % metadata_filename,
          '--output_path=%s' % output_dir,
-         '--transforms_config_file=%s' % transforms_config_file,
+         '--schema_file=%s' % schema_file,
```
qimingj: Can preprocessing save the schema file into preprocessed_dir so that we don't have to specify it here?

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
```diff
@@ -230,16 +222,17 @@ def local_train(preprocessed_dir, transforms_config_file, output_dir,
   print('Local training done.')
 
 
-def cloud_train(preprocessed_dir, transforms_config_file, output_dir,
-                staging_bucket,
+def cloud_train(preprocessed_dir, schema_file, transforms_file, output_dir,
+                model_type, staging_bucket,
```
qimingj: I hope staging-bucket can be optional. For example, use output_dir.

solutionbox/structured_data/datalab_solutions/structured_data/test/e2e_functions.py
```diff
@@ -88,35 +95,38 @@ def run_preprocess(output_dir, csv_filename, config_filename,
...
+         '--schema_file=%s' % schema_filename,
+         '--transforms_file=%s' % transforms_filename,
          '--max_steps=2500'] + extra_args
   print('Current working directoyr: %s' % os.getcwd())
   print('Going to run command: %s' % ' '.join(cmd))
-  sp = subprocess.Popen(' '.join(cmd), shell=True, stderr=subprocess.PIPE) #open(os.devnull, 'wb'))
+  sp = subprocess.Popen(' '.join(cmd), shell=True, stderr=subprocess.PIPE) 
   _, err = sp.communicate()
   err = err.splitlines()
   print 'last line'
```
qimingj: This is your own debugging code right?

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
```diff
@@ -328,23 +320,144 @@ def cloud_train(preprocessed_dir, transforms_config_file, output_dir,
...
-def local_batch_predict():
-  """Not Implemented Yet."""
-  print('local_batch_predict')
+  if (not prediction_input_file.startswith('gs://') or
+      not output_dir.startswith('gs://')):
+    print('ERROR: prediction_input_file and output_dir must point to a '
+          'location on GCS.')
+    return
 
+  cmd = ['gcloud beta ml jobs submit prediction %s' % job_name,
```
qimingj: Consider using https://github.com/googledatalab/pydatalab/blob/master/datalab/mlalpha/_cloud_predictor.py. Gcloud works but we may have to monitor the external process. Also, we want to pass the datalab header (line 84 in _cloud_predictor.py). Same for training.

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
```diff
@@ -184,24 +175,23 @@ def cloud_preprocess(input_file_path, output_dir, transforms_config_file,
...
+                model_type,
                 layer_sizes=None, max_steps=None):
   """Train model locally.
   Args:
     preprocessed_dir: The output directory from preprocessing. Must contain
         files named features_train*.tfrecord.gz, features_eval*.tfrecord.gz,
         and metadata.json. Can be local or GCS path.
-    transforms_config_file: File path to the config file.
+    schema_file: Same file used for preprocessing
+    transforms_file: File describing transforms to perform on the data.
```
qimingj: missing model_type docstring? Same for cloud_train.

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
qimingj: It should not block this change. We can do it separately since it will be a big change.

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
```diff
@@ -107,29 +104,26 @@ def local_preprocess(input_file_path, output_dir, transforms_config_file,
         files. Preprocessing will automatically slip the data into three sets
         for training, evaluation, and testing. Can be local or GCS path.
     output_dir: The output directory to use; can be local or GCS path.
-    transforms_config_file: File path to the config file.
+    schema_file: File path to the schema file.
     train_percent: Int in range [0, 100].
     eval_percent: Int in range [0, 100].
     train_percent: Int in range [0, 100].
```
brandondutra: done

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
brandondutra: done

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
brandondutra: good idea. done.

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
brandondutra: stage-bucket has to be a real bucket (gs://bucket) and not a path (gs://bucket/folder). I would rather ask the user, otherwise they might think "why, are there all these files in my bucket, who made them?"

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
brandondutra: good point. I'll do it in a future pull request.

solutionbox/structured_data/datalab_solutions/structured_data/test/e2e_functions.py
brandondutra: yes. This is part of the tests, so the test does not print all the training output, just the last line that has the loss values.

solutionbox/structured_data/datalab_solutions/structured_data/_package.py
qimingj: It's the same thing with DataFlow which requires a tmp/stage location. It feels like asking for a temp GCS dir. Personally I think it is Okay to default to output folder, but up to you.

brandondutra: Very true, but for dataflow, the stage location can be any gcs path (gs://bucket/my/path/tmp/stage), So it's easy to say that we can make the stage location just "output_dir/tmp/stage".

However, for training (via gcloud?) the stage location has to be the root bucket (so gs://bucket/my/path/tmp/stage does not work). Once they fix that issue, I want to move over to just using the output folder.

qimingj: Actually if you call API for training, it seems not asking for a staging bucket...

PR #141, opened 2017-01-28
Remove bq.Query.results(), remove SQL tokenization

yebrahim:
- Rename `%bigquery` to `%bq` in the google.datalab package
- Remove `data.tokenize` function. SQL is now passed to BigQuery without parsing, which means UDFs and external datasources will not work just yet. Simple `$`-prefixed parameters are still expanded.
- Add missing `google.` to all imports and module calls
- Disable UDF and external datasource tests for now, more changes are coming to their syntax


google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -279,10 +279,10 @@ def _get_query_argument(args, cell, env):
 
 
 def _sample_cell(args, cell_body):
-  """Implements the bigquery sample cell magic for ipython notebooks.
+  """Implements the bq sample cell magic for ipython notebooks.
```
qimingj: Maybe it should be "Implements the BigQuery sample" so it is consistent with others?

yebrahim: Done.

PR #142, opened 2017-01-29
Resolve issue where datalab build fails when installing pyDatalab

parthea:
This PR fixes issue googledatalab/datalab#1160 in my local setup with the following changes:

- In setup.py, extend setuputils install to include additional pip installations for tensorflow and cloud ml similar to https://docs.python.org/2/distutils/extending.html.
- In setup.py, `platform.linux_distribution()[0]` was returning `'debian'` on Ubuntu. On my machine, `pip._vendor.distro.linux_distribution()[0]` provides the correct distribution `'Ubuntu'`.

```
>>> import platform
>>> platform.linux_distribution()[0]
'debian'
```

```
>>> from pip._vendor import distro
>>> distro.linux_distribution()[0]
u'Ubuntu'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,13 +15,39 @@
 import datetime
 import sys
 from setuptools import setup
+from setuptools.command.install import install
 
 if sys.version_info[0] == 2:
-  import platform
   import pip
 
 minor = datetime.datetime.now().strftime("%y%m%d%H%M")
 version = '0.1.' + minor
+
+
+class InstallPipPackages(install):
+    def run(self):
+        install.run(self)
+        # for python2 only, install tensorflow and cloudml
+        if sys.version_info[0] == 2:
+          tensorflow_path = None
+          if sys.platform.startswith("darwin"):
+            tensorflow_path = 'https://storage.googleapis.com/tensorflow/mac/cpu/tensorflow-0.11.0-py2-none-any.whl'
+          elif sys.platform.startswith("linux"):
+            from pip._vendor import distro
+            if distro.linux_distribution()[0].startswith('Ubuntu'):
+              tensorflow_path = 'https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.11.0-cp27-none-linux_x86_64.whl'
+            elif distro.linux_distribution()[0].startswith('Debian'):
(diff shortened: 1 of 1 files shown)


yebrahim:
Interesting. Why does this fix it though? I don't see any differences between this and installing TF and CloudML separately in the setup script.

parthea:
> Why does this fix it though? I don't see any differences between this and installing TF and CloudML separately in the setup script.

Although the build failure doesn't appear when I apply the changes in this PR, I wasn't able to find/explain the root cause. I'm going to close this PR so we can dig into this a bit more.

I prefer `pip._vendor.distro.linux_distribution()[0]` over `platform.linux_distribution()[0]` based on the output of these 2 commands on my Ubuntu laptop. The build failure exists regardless of whether the Linux distribution is detected correctly.

parthea:
I noticed that the build failure doesn't appear when I comment out the following line in setup.py. Hopefully, this is a clue to help find the root cause.

```
  pip.main(['install', 'https://storage.googleapis.com/cloud-ml/sdk/cloudml.latest.tar.gz'])
```

https://github.com/googledatalab/pydatalab/blob/master/setup.py#L132

yebrahim:
Honestly I'm not sure if it's worth spending more time investigating this, I think we need a workaround. 

I also noticed the error goes away if we use an older version of pip (<9), and I've seen some other threads about this being a bug in pip 9+, but I still can't put my hands on the exact root cause. Either fix is fine to me. I'll reopen your PR so we keep our options open.

yebrahim:
Can't reopen, you deleted your branch. :)

parthea:
Done.

It would be interesting to see if the issue re-appears with this workaround. I would be happy to dig into this issue further if we are interested to find the root cause. I am really curious about the root cause as well.

yebrahim:
Ok so CloudML SDK released a new version on Friday, and this is what's causing the break. I'm checking to see what our options are, but for the time being, I don't think further digging will get us anywhere. :)

yebrahim:
I just sent a workaround to downgrade pip: https://github.com/googledatalab/datalab/pull/1162, since this one is breaking tests for some reason. I suspect the reason is CloudML isn't installing properly, as I couldn't get it to install inside the Datalab container. I'll close this PR. Thanks for digging into this @parthea.

PR #144, opened 2017-01-31
Remove tensorflow and CloudML SDK from setup.py

qimingj:
There are cases that users want to use BigQuery APIs from pydatalab but not ML stuff, so having a hard dependency on it is not ideal.

### DIFF
--- a/datalab/kernel/__init__.py
+++ b/datalab/kernel/__init__.py
@@ -31,13 +31,20 @@ import datalab.context as _context
 import datalab.bigquery.commands
 import datalab.context.commands
 import datalab.data.commands
+import datalab.stackdriver.commands
+import datalab.storage.commands
+import datalab.utils.commands
+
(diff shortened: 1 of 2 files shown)


qimingj:
PTAL.

datalab/kernel/__init__.py
```diff
@@ -31,7 +31,10 @@
 import datalab.bigquery.commands
 import datalab.context.commands
 import datalab.data.commands
-import datalab.mlalpha.commands
+try:
+  import datalab.mlalpha.commands
+except:
+  print('TensorFlow and CloudML SDK are required.')
```
yebrahim: Can we elaborate more on this, mainly make it clear what will be disabled as a result. Maybe also adding links to install TF and CloudML would be a good idea, I think a long comment is good here.

qimingj: Done.

PR #151, opened 2017-02-01
Remove old DataSet implementation. Create new DataSets.

qimingj:
The new Dataset will be used as data source for packages. All DataSets will be capable of sampling to DataFrame, so feature exploration can be done with other libraries.

### DIFF
--- a/datalab/mlalpha/_dataset.py
+++ b/datalab/mlalpha/_dataset.py
@@ -1,4 +1,4 @@
-# Copyright 2016 Google Inc. All rights reserved.
+# Copyright 2017 Google Inc. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


datalab/mlalpha/_dataset.py
```diff
@@ -11,407 +11,120 @@
...
+    row_counts = []
+    for file in self._files:
+      with ml.util._file.open_local_or_gcs(file, 'r') as f:
+        num_lines = sum(1 for line in f)
+        row_total_count += num_lines
+        row_counts.append(num_lines)
+
+    names = None
+    dtype = None
+    if self._schema:
```
brandondutra: how could self._schema be none? Maybe just use an assert?

datalab/mlalpha/_dataset.py
```diff
@@ -11,407 +11,120 @@
...
+    """Samples data into a Pandas DataFrame. Note that it calls BigQuery so it will
+       incur cost.
+    Args:
+      n: number of sampled counts. Note that the number of counts returned is approximated.
+    Returns:
+      A dataframe containing sampled data.
+    """
+    total = bq.Query('select count(*) from (%s)' % self._sql).results()[0].values()[0]
+    sampling = bq.Sampling.random(n*100.0/float(total))
+    sample = bq.Query(self._sql).sample(sampling=sampling)
```
brandondutra: what if you over sample, and then do a "select * from () limit n" statement?

datalab/mlalpha/_dataset.py
qimingj: It can be None. I choose to make it optional in case the package knows the schema beforehand (like in inception).

datalab/mlalpha/_dataset.py
qimingj: Done.

PR #155, opened 2017-02-02
Inception package improvements

qimingj:
- It takes DataSets as input instead of CSV files. It also supports BigQuery source now.
- Changes to make latest DataFlow and TensorFlow happy.
- Changes in preprocessing to remove partial support for multiple labels.
- Other minor improments.

### DIFF
--- a/solutionbox/inception/datalab_solutions/inception/_cloud.py
+++ b/solutionbox/inception/datalab_solutions/inception/_cloud.py
@@ -20,10 +20,10 @@ import apache_beam as beam
 import base64
 import collections
 import datetime
-from googleapiclient import discovery
 import google.cloud.ml as ml
 import logging
 import os
+
 
 from . import _model
 from . import _preprocess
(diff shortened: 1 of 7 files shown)


solutionbox/inception/datalab_solutions/inception/_cloud.py
```diff
@@ -20,18 +20,18 @@
...
 
+
 from . import _model
 from . import _preprocess
 from . import _trainer
 from . import _util
 
 
-_TF_GS_URL= 'gs://cloud-datalab/deploy/tf/tensorflow-0.12.0rc0-cp27-none-linux_x86_64.whl'
+_TF_GS_URL= 'gs://cloud-datalab/deploy/tf/tensorflow-0.12.0rc1-cp27-none-linux_x86_64.whl'
```
chmeyers: Why aren't we using an official TF release url here?

qimingj: We could. But I am not sure how long they will keep these releases there. It seems better to store dependencies in our own bucket, as much as possible?

PR #158, opened 2017-02-03
New %%bq query magic

yebrahim:
Reworked %bq command, this PR fixes some missing arguments.

The bulk of this PR is in the `_query_cell` function, which parses the magic call into individual API calls, as well as `_create_query_parser` function, which includes the the argparse calls.

### DIFF
--- a/google/datalab/bigquery/_sampling.py
+++ b/google/datalab/bigquery/_sampling.py
@@ -111,3 +111,25 @@ class Sampling(object):
         sql = '%s LIMIT %d' % (sql, count)
       return sql
     return _random_sampling
+
+  @staticmethod
+  def _auto(method, fields, count, percent, key_field, ascending):
+    """Construct a Sorting object according to the provided sampling technique, provided all
+    its needed fields are passed as arguments
+
+    Args:
+      method: sampling strategy, one of the method names above in this module
+      fields: an optional list of field names to retrieve.
+      count: maximum number of rows to limit the sampled results to.
+      percent: the percentage of the resulting hashes to select if using hashed sampling
+      key_field: the name of the field to sort the rows by or use for hashing
+      ascending: whether to sort in ascending direction or not.
+    """
+    if method == 'limit':
+      return Sampling.default(fields=fields, count=count)
+    elif method == 'random':
+      return Sampling.random(fields=fields, percent=percent, count=count)
+    elif method == 'hashed':
+      return Sampling.hashed(fields=fields, field_name=key_field, percent=percent, count=count)
+    elif method == 'sorted':
+      return Sampling.sorted(fields=fields, field_name=key_field, ascending=ascending, count=count)
--- a/google/datalab/bigquery/_sampling.py
+++ b/google/datalab/bigquery/_sampling.py
@@ -65,6 +65,8 @@ class Sampling(object):
     Returns:
       A sampling function that can be applied to get the initial few rows.
     """
+    if field_name is None:
+      raise Exception('Sort field must be specified')
     direction = '' if ascending else ' DESC'
     projection = Sampling._create_projection(fields)
     return lambda sql: 'SELECT %s FROM (%s) ORDER BY %s%s LIMIT %d' % (projection, sql, field_name,
@@ -81,6 +83,8 @@ class Sampling(object):
     Returns:
       A sampling function that can be applied to get a hash-based sampling.
     """
+    if field_name is None:
+      raise Exception('Hash field must be specified')
     def _hashed_sampling(sql):
       projection = Sampling._create_projection(fields)
       sql = 'SELECT %s FROM (%s) WHERE ABS(HASH(%s)) %% 100 < %d' % \
@@ -114,16 +118,20 @@ class Sampling(object):
 
   @staticmethod
   def _auto(method, fields, count, percent, key_field, ascending):
-    """Construct a Sorting object according to the provided sampling technique, provided all
+    """Construct a sampling function according to the provided sampling technique, provided all
     its needed fields are passed as arguments
 
     Args:
-      method: sampling strategy, one of the method names above in this module
+      method: one of the supported sampling methods: {limit,random,hashed,sorted}
       fields: an optional list of field names to retrieve.
       count: maximum number of rows to limit the sampled results to.
       percent: the percentage of the resulting hashes to select if using hashed sampling
       key_field: the name of the field to sort the rows by or use for hashing
       ascending: whether to sort in ascending direction or not.
+    Returns:
+      A sampling function using the provided arguments
+    Raises:
+      Exception if an unsupported mathod name is passed
     """
     if method == 'limit':
       return Sampling.default(fields=fields, count=count)
@@ -133,3 +141,6 @@ class Sampling(object):
       return Sampling.hashed(fields=fields, field_name=key_field, percent=percent, count=count)
     elif method == 'sorted':
       return Sampling.sorted(fields=fields, field_name=key_field, ascending=ascending, count=count)
+    else:
+      raise Exception('Unsupported sampling method: %s' % method)
+
--- a/google/datalab/bigquery/commands/_bigquery.py
+++ b/google/datalab/bigquery/commands/_bigquery.py
@@ -98,74 +98,6 @@ def _create_sample_subparser(parser):
   return sample_parser
 
 
-def _create_udf_subparser(parser):
-  udf_parser = parser.subcommand('udf', 'Create a named Javascript BigQuery UDF')
-  udf_parser.add_argument('-m', '--module', help='The name for this UDF')
-  return udf_parser
-
-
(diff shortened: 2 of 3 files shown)


google/datalab/bigquery/_sampling.py
```diff
@@ -111,3 +111,25 @@ def _random_sampling(sql):
...
+    """Construct a Sorting object according to the provided sampling technique, provided all
+    its needed fields are passed as arguments
+
+    Args:
+      method: sampling strategy, one of the method names above in this module
+      fields: an optional list of field names to retrieve.
+      count: maximum number of rows to limit the sampled results to.
+      percent: the percentage of the resulting hashes to select if using hashed sampling
+      key_field: the name of the field to sort the rows by or use for hashing
+      ascending: whether to sort in ascending direction or not.
```
ojarjur: This needs a "Returns:" block in the docstring.

google/datalab/bigquery/_sampling.py
```diff
@@ -111,3 +111,25 @@ def _random_sampling(sql):
         sql = '%s LIMIT %d' % (sql, count)
       return sql
     return _random_sampling
+
+  @staticmethod
+  def _auto(method, fields, count, percent, key_field, ascending):
+    """Construct a Sorting object according to the provided sampling technique, provided all
```
ojarjur: Change "Sorting object" to either "sampling function" or "sampling strategy" to be consistent with how the rest of this file refers to the type of value returned from this method.

google/datalab/bigquery/_sampling.py
```diff
@@ -111,3 +111,25 @@ def _random_sampling(sql):
...
+      ascending: whether to sort in ascending direction or not.
+    """
+    if method == 'limit':
+      return Sampling.default(fields=fields, count=count)
+    elif method == 'random':
+      return Sampling.random(fields=fields, percent=percent, count=count)
+    elif method == 'hashed':
+      return Sampling.hashed(fields=fields, field_name=key_field, percent=percent, count=count)
+    elif method == 'sorted':
+      return Sampling.sorted(fields=fields, field_name=key_field, ascending=ascending, count=count)
```
ojarjur: This should have a final 'else' block to define how to handle a method that does not fit in the supported set.

I would suggest making that raise an exception, and adding a "Raises:" section to the docstring.

google/datalab/bigquery/_sampling.py
```diff
@@ -111,3 +111,25 @@ def _random_sampling(sql):
...
       return sql
     return _random_sampling
+
+  @staticmethod
+  def _auto(method, fields, count, percent, key_field, ascending):
+    """Construct a Sorting object according to the provided sampling technique, provided all
+    its needed fields are passed as arguments
+
+    Args:
+      method: sampling strategy, one of the method names above in this module
```
ojarjur: This isn't quite right. For instance, `limit` doesn't match `default`.

I would just explicitly list the supported methods.

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -400,154 +402,6 @@ def _delete_cell(args, _):
...
-  query = _get_query_argument(args, cell_body, google.datalab.utils.commands.notebook_environment())
-
-  if args['verbose']:
-    print(query.sql)
-  result = query.execute_dry_run(dialect=args['dialect'], billing_tier=args['billing'])
-  return google.datalab.bigquery._query_stats.QueryStats(total_bytes=result['totalBytesProcessed'],
-                                              is_cached=result['cacheHit'])
-
-
-def _udf_cell(args, js):
```
ojarjur: I think we should keep the current udf and pipeline code until the replacement is ready.

I.E. Add the new functionality and remove the old in one atomic change.

That will make it much easier for us to compare the new bits vs. the old to make sure we don't forget anything in the transition.

tests/kernel/bigquery_tests.py
```diff
@@ -37,32 +37,32 @@ def noop_decorator(func):
...
-    mock_default_context.return_value = TestCases._create_context()
-    mock_notebook_environment.return_value = env
-    google.datalab.bigquery.commands._bigquery._udf_cell({'module': 'word_filter'}, cell_body)
-    udf = env['word_filter']
-    self.assertIsNotNone(udf)
-    self.assertEquals('word_filter', udf._name)
-    self.assertEquals([('word', 'string'), ('corpus', 'string'), ('count', 'integer')],
-                      udf._outputs)
-    self.assertEquals(cell_body, udf._implementation)
+  #@mock.patch('google.datalab.utils.commands.notebook_environment')
```
ojarjur: Why are we keeping the commented-out test code?

I think we should just keep it (along with the udf implementation), until the new udf functionality is in.

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -770,6 +593,96 @@ def _add_command(parser, subparser_fn, handler, cell_required=False, cell_prohib
...
+  execute_parser.add_argument('-m', '--mode', help='The table creation mode', default='create',
+                              choices=['create', 'append', 'overwrite'])
+  execute_parser.add_argument('-l', '--large', help='Whether to allow large results',
+                              action='store_true')
+  execute_parser.add_argument('-q', '--query', help='The name of query to run')
+  execute_parser.add_argument('-t', '--table', help='Target table name')
+  execute_parser.add_argument('--to-dataframe', help='Convert the result into a dataframe',
+                              action='store_true')
+  execute_parser.add_argument('--dataframe-start-row', help='Row of the table to start the ' +
+                              'dataframe export', default=0)
```
qimingj: Should default here be None since you check for None in the code above?

google/datalab/bigquery/commands/_bigquery.py
yebrahim: Sure

tests/kernel/bigquery_tests.py
yebrahim: Done

google/datalab/bigquery/commands/_bigquery.py
yebrahim: Good point, we should minimize the number of places the default is passed

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -279,63 +280,64 @@ def _get_query_argument(args, cell, env):
...
+    # if an execute was performed, a QueryJob is returned, which has a .results property
+    # otherwise, it's a base Job object, .result() should be called
+    return r.results if hasattr(r, 'results') else r.result()
+
+  elif args['command'] == 'sample':
+    # parse comma-separated list of fields
+    fields = args['fields'].split(',') if args['fields'] else None
+    count = int(args['count']) if args['count'] else None
+    sampling=Sampling._auto(method=args['method'], fields=fields, count=count,
+                            percent=int(args['percent']), key_field=args['key_field'],
```
yebrahim: Good catch

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -279,63 +280,64 @@ def _get_query_argument(args, cell, env):
...
+
+    r = query.execute(output_options, dialect=args['dialect'], billing_tier=args['billing'])
+    # if an execute was performed, a QueryJob is returned, which has a .results property
+    # otherwise, it's a base Job object, .result() should be called
+    return r.results if hasattr(r, 'results') else r.result()
+
+  elif args['command'] == 'sample':
+    # parse comma-separated list of fields
+    fields = args['fields'].split(',') if args['fields'] else None
+    count = int(args['count']) if args['count'] else None
```
yebrahim: It can't. It can only be '0', which will pass the if test

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -279,63 +280,64 @@ def _get_query_argument(args, cell, env):
...
+    elif args['command'] == 'extract':
+      output_options = QueryOutput.file(path=args['path'], format=args['format'],
+                                        csv_delimiter=args['delimiter'],
+                                        csv_header=args['header'], compress=args['compress'],
+                                        use_cache=not args['nocache'])
+
+    r = query.execute(output_options, dialect=args['dialect'], billing_tier=args['billing'])
+    # if an execute was performed, a QueryJob is returned, which has a .results property
+    # otherwise, it's a base Job object, .result() should be called
+    return r.results if hasattr(r, 'results') else r.result()
```
yebrahim: It does, methods are attributes

qimingj: So based on your comments, if it returns a base Job object, even though it does not have .results field, hasattr(r, 'results') will be True because it has .results() method, and the function will return r.results which is a method?

yebrahim: The method and property names are different, there's an extra 's'. :)
I know, I have a comment to change that.

PR #159, opened 2017-02-03
Move confusion matrix from %%ml to library.

qimingj:
This is part of efforts to move %%ml magic stuff to library to provide a consistent experience (python only).

### DIFF
--- a/datalab/mlalpha/_confusion_matrix.py
+++ b/datalab/mlalpha/_confusion_matrix.py
@@ -42,6 +42,7 @@ class ConfusionMatrix(object):
       input_csv: Path to a Csv file (with no header). Can be local or GCS path.
       headers: Csv headers. If present, it must include 'target' and 'predicted'.
       schema_file: Path to a JSON file containing BigQuery schema. Used if "headers" is None.
+          If present, it must include 'target' and 'predicted' columns.
     Returns:
       A ConfusionMatrix that can be plotted.
     Raises:
--- a/datalab/mlalpha/_confusion_matrix.py
+++ b/datalab/mlalpha/_confusion_matrix.py
@@ -11,57 +11,99 @@
 # the License.
 
 
-from plotly.offline import iplot
+import google.cloud.ml as ml
+import numpy as np
+import json
+import matplotlib.pyplot as plt
+import pandas as pd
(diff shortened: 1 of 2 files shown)


datalab/mlalpha/_confusion_matrix.py
```diff
@@ -11,57 +11,99 @@
...
+  @staticmethod
+  def from_csv(input_csv, headers=None, schema_file=None):
+    """Create a ConfusionMatrix from a csv file.
     Args:
-      predicted_labels: a list of predicted labels.
-      true_labels: a list of true labels.
-      counts: a list of count for each (predicted, true) combination.
+      input_csv: Path to a Csv file (with no header). Can be local or GCS path.
+      headers: Csv headers. If present, it must include 'target' and 'predicted'.
+      schema_file: Path to a JSON file containing BigQuery schema. Used if "headers" is None.
```
brandondutra: Add "If present, it must include 'target' and 'predicted'." to the schema_file line too or just say there has to be two columns named target and predicted before the args list.

datalab/mlalpha/_confusion_matrix.py
```diff
@@ -11,57 +11,99 @@
...
-      data.append(predicts_for_current_true_label)
-    self._data = data
+
+    query, _ = data.SqlModule.get_sql_statement_with_environment(sql, {})
+    sql = ('SELECT target, predicted, count(*) as count FROM (%s) group by target, predicted'
+        % query.sql)
+    df = bq.Query(sql).results().to_dataframe()
+    labels = sorted(set(df['target']) | set(df['predicted']))
+    labels_count = len(labels)
+    df['target'] = [labels.index(x) for x in df['target']]
```
brandondutra: why don't we just call confusion_matrix() like above?

qimingj: We cannot. Scikit-learn's confusion-matrix constructor only accepts a list of "target" and a list of "predicted" (so it doesn't work with large data). It does not accept aggregated data.

datalab/mlalpha/_confusion_matrix.py
qimingj: Done

PR #166, opened 2017-02-06
Google datalab chmeyers

chmeyers:
Making Job a top level class and making all derived Job classes hidden.


google/datalab/bigquery/_view.py
```diff
@@ -186,7 +186,7 @@ def results(self, use_cache=True, dialect=None, billing_tier=None):
     """
     output_options = QueryOutput.table(use_cache=use_cache)
     return self._materialization.execute(output_options, dialect=dialect,
-                                               billing_tier=billing_tier).results
+                                               billing_tier=billing_tier).result()
```
yebrahim: Otherwise, we'd be calling the base class' `result()` method, which isn't what we want to do here.
The `QueryJob` and `Job` classes need clear resolution in my opinion, they're too confusing side by side. I suggest changing `QueryJob.results` into `QueryJob.query_result()`, and `Job.result()` into `Job.job_result()` so that either can be called on the inherited type.
Another option would be to keep only one type `Job` that's aware of both async operations as well as GCP operations, and its `.results()` returns the GCP operation's results, or falls back to the async operation's.

chmeyers: To match the base class, I made QueryJob.result a non-property.

How sure are you that it would call the base classes result()?  Not a huge python expert, but from my tests and what I could find the derived class' method overwrites the base class method.

result() doesn't have a defined return type, it is just whatever type the future returns, which is a table when the job is a query.

yebrahim: Oh I misread the diff, didn't notice you made it a non-property. Yes the inherited type's method will override the base's.

The only issue now is we don't have access to the GCP operation's result. I think this is fine for now since we don't see any scenarios that need this off the top of my head.

Btw, are there any more `.results` references that need fixing?? I think at least `_query.py` and `_bigquery.py` will have some.

chmeyers: I did miss _query.py and _bigquery.py.  Fixed those and did a sweep but didn't find any other places.

yebrahim: We have a lot of work to do on test coverage. :)

PR #168, opened 2017-02-07
Move all references to dialect/billing-tier into global Context object

ojarjur:
This change switches all modules under 'google/datalab' to using the mutable `config` object inside of the `context` object for determining the dialect and billing tier for BigQuery API calls.


google/datalab/bigquery/_api.py
```diff
@@ -40,13 +40,22 @@ def __init__(self, context):
...
 
   @property
   def project_id(self):
     """The project_id associated with this API client."""
-    return self._project_id
+    return self._context.project_id
+
+  @property
+  def bigquery_dialect(self):
+    """The project_id associated with this API client."""
```
yebrahim: docstring

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -250,7 +250,26 @@ def _create_load_subparser(parser):
...
+    args: the dictionary of magic arguments.
+  Returns:
+    A Context based on the current default context, but with any explicitly specified
+      arguments overriding the defaults.
+  """
+  context = google.datalab.Context.default()
+  dialect_arg = args.get('dialect', None)
+  billing_tier_arg = args.get('billing', None)
+  if dialect_arg:
+    context.config['bigquery_dialect'] = dialect_arg
```
yebrahim: doesn't this overwrite the global context for all future executions too? i think we need to modify the `.default()` method to make it return a new copy on every call, and add a `.global_context()` instead of it.

google/datalab/bigquery/_api.py
ojarjur: Fixed.

google/datalab/bigquery/commands/_bigquery.py
ojarjur: You're right, this would have overwritten the global default.

I've changed just this method to return a new Context rather than using the global default.

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -250,23 +250,31 @@ def _create_load_subparser(parser):
...
+  global_default_context = google.datalab.Context.default()
+  config = {}
+  for key in global_default_context.config:
+    config[key] = global_default_context.config[key]
+
   dialect_arg = args.get('dialect', None)
   billing_tier_arg = args.get('billing', None)
   if dialect_arg:
-    context.config['bigquery_dialect'] = dialect_arg
+    config['bigquery_dialect'] = dialect_arg
```
yebrahim: why pick these two config values individually? we might add things in the future, can't we just copy all the keys in `args` over the ones in `config`?

ojarjur: > can't we just copy all the keys in args over the ones in config

Unfortunately, the keys don't match.

E.G. `bigquery_billing_tier` vs. just `billing`

PR #169, opened 2017-02-07
A util function to repackage and copy the package to staging location.

qimingj:
A util function to repackage and copy the package to staging location, so in packages we can use the staging URL as package URL in cloud training.


datalab/mlalpha/_util.py
```diff
@@ -0,0 +1,59 @@
...
+  previous_cwd = os.getcwd()
+  os.chdir(package_dir)
+  try:
+    # Repackage. In Datalab the source files of pydatalab is kept.
+    setup_py = os.path.join(package_dir, 'setup.py')
+    sdist = ['python', setup_py, 'sdist', '--format=gztar', '-d', tempdir]
+    subprocess.check_call(sdist)
+
+    # Copy to GCS.
+    dest = os.path.join(out_dir, 'staging', package_name + '.tar.gz')
```
brandondutra: 1) Does this work with version names "inception-1.0.0"?
2) This would only work for us, not for any other package author, as other package authors may not keep their source in /datalab/lib/pydatalab/, right?

I was expecting a function that takes a file path to the folder containing the setup.py file. I don't think we should make assumptions on where the package is. Thoughts?

qimingj: Changed implementation so that:

1. it takes a setup.py file path (package is responsible to install it), and a package root dir (which may not contain setup.py, such as when it passes in parent of parent folder of  __file__). So we don't have dependency on /datalab/lib/pydatalab/.

2. It works with version names because there should be only one "*.tar.gz" file in the temp dir.

datalab/mlalpha/_util.py
```diff
@@ -0,0 +1,65 @@
...
+  """
+  if not out_dir.startswith('gs://'):
+    raise ValueError('Output needs to be a GCS path.')
+
+  dest_setup_py = os.path.join(package_root_dir, 'setup.py')
+  # setuptools requires a "setup.py" in the current dir. So check if
+  # there is existing setup.py. If so, back it up.
+  if os.path.isfile(dest_setup_py):
+    os.rename(dest_setup_py, dest_setup_py + '._bak_')
+  shutil.copyfile(setup_py, dest_setup_py)
```
brandondutra: also check that setup_py exists?

datalab/mlalpha/_util.py
```diff
@@ -0,0 +1,65 @@
...
+
+import datetime
+import os
+import shutil
+import subprocess
+import tempfile
+
+
+def package_and_copy(package_root_dir, setup_py, out_dir):
+  """Repackage an CloudML package and copy it to a staging dir.
```
brandondutra: maybe allow caller to set output name? If I run 10 training jobs, why should we make 10 copies?

qimingj: Makes sense. Done!

datalab/mlalpha/_util.py
qimingj: Done.

PR #171, opened 2017-02-07
Add %%bq query to build BigQuery query objects incrementally

yebrahim:
This PR adds the `%%bq query` which can be used to build BigQuery query modules that can consume other, similar modules, or UDF objects.

The change also adds recursive subquery expansion. It does *not* cover circular subquery reference, this is rather delegated to the BigQuery service.


google/datalab/bigquery/_query.py
```diff
@@ -91,6 +107,41 @@ def __init__(self, sql, context=None, values=None, udfs=None, data_sources=None,
           raise Exception('Referenced external table %s has no known schema' % name)
         self._external_tables[name] = table._to_query_json()
 
+  @property
+  def expanded_sql(self):
```
ojarjur: We really need to add a unit test for this.

yebrahim: I agree, we have a lot of work to do on unit tests.

google/datalab/bigquery/_query.py
```diff
@@ -65,20 +67,34 @@ def __init__(self, sql, context=None, values=None, udfs=None, data_sources=None,
...
 
     if data_sources is None:
       data_sources = {}
 
     self._code = None
     self._imports = []
-    if values is None:
-      values = kwargs
+    if self._values is None:
+      self._values = kwargs
```
chmeyers: This is going to conflict with my change. :)

Just make it self._values = {} and remove kwargs above if you want to reduce the conflict.

google/datalab/bigquery/_query.py
```diff
@@ -91,13 +107,52 @@ def __init__(self, sql, context=None, values=None, udfs=None, data_sources=None,
...
+    def _recurse_subqueries(query):
+      """Recursively scan subqueries and add their pieces to global scope udfs and subqueries
+      """
+      if query._subqueries:
+        subqueries.update(query._subqueries)
+      if query._udfs:
+        udfs.update(set(query._udfs))
+      if query._subqueries:
+        for subquery in query._subqueries:
+          subquery = self._values[subquery]
```
chmeyers: Why not just _recurse_subqueries(self._values[subquery]) ?

google/datalab/bigquery/_query.py
yebrahim: Done

google/datalab/bigquery/_query.py
yebrahim: Done

PR #174, opened 2017-02-08
Removing the remaining obsolete classes/methods from the refactoring list.

chmeyers:


google/datalab/bigquery/_query.py
```diff
@@ -34,9 +34,8 @@ class Query(object):
   This object can be used to execute SQL queries and retrieve results.
   """
 
-  def __init__(self, sql, context=None, values=None, udfs=None, data_sources=None, **kwargs):
+  def __init__(self, sql, context=None, values={}, udfs=None, data_sources=None):
```
yebrahim: > values={}
this is likely to be problematic if `values` is ever changed. please use `None` and assign an empty dictionary instead

google/datalab/bigquery/_table.py
```diff
@@ -636,46 +636,6 @@ def to_dataframe(self, start_row=0, max_rows=None):
     ordered_fields = [field.name for field in self.schema]
     return df[ordered_fields] if df is not None else pandas.DataFrame()
 
-  def to_file(self, destination, format='csv', csv_delimiter=',', csv_header=True):
```
yebrahim: note this has the actual extract functionality that the `Query` object uses. if we don't want to expose it we should just make it private, but we can't remove it altogether.

google/datalab/storage/_bucket.py
```diff
@@ -235,18 +235,6 @@ def contains(self, name):
...
-    """Creates a new bucket.
-
-    Args:
-      name: a unique name for the new bucket.
-    Returns:
-      The newly created bucket.
-    Raises:
-      Exception if there was an error creating the bucket.
-    """
-    return Bucket(name, context=self._context).create(self._project_id)
```
yebrahim: are we losing `context` setting in this case? is it needed for storage?

google/datalab/bigquery/_query.py
chmeyers: Done

google/datalab/storage/_bucket.py
chmeyers: We expect users to use Bucket().create() directly now, so they send the context into that.

google/datalab/bigquery/_table.py
chmeyers: Made it private and updated _query.py.  Why did we decide to remove to_file() but not to_dataframe()?

yebrahim: I think we should remove both

chmeyers: Done.  make to_dataframe private as well.

tests/kernel/sql_tests.py
```diff
@@ -128,10 +128,6 @@ def test_arguments(self, mock_default_context):
     sql = google.datalab.bigquery.Query(query, values={}).sql
     self.assertEquals('SELECT * FROM [publicdata:samples.shakespeare]\n' +
                       'WHERE word IN ("thus", "forsooth")\nLIMIT 10', sql)
-    # As above but with overrides, using list
```
yebrahim: why remove the tests?

chmeyers: There was one test that used kwargs, and one that used values.  Since we are removing the kwargs options, I removed the kwargs test.  The values test still covers the remaining behavior.

PR #175, opened 2017-02-08
Improve inception package so there is no need to have an GCS copy.

qimingj:
Improve inception package so there is no need to keep a GCS copy of the package. Instead cloud training and cloud preprocessing will repackage it from local installation and upload it to staging on the fly.


solutionbox/inception/setup.py
```diff
@@ -25,6 +25,9 @@
     'datalab_solutions',
     'datalab_solutions.inception',
   ],
+  # setup.py needs to be deployed so it can be repackaged from local installation for cloud run.
+  data_files=[('/datalab/packages_setup/inception', ['setup.py'])],
```
brandondutra: can it be relative? This will not work outside the vm, right?

qimingj: It can be relative, but AFAIK it is relative to sys.prefix (which also depends on what user it runs under). This should work with Linux/Mac, if user installs it with root.

PR #178, opened 2017-02-09
Add CloudTrainingConfig namedtuple to wrap cloud training configurations

qimingj:


datalab/mlalpha/_cloud_training_config.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections import namedtuple
+
+_CloudTrainingConfig = namedtuple("CloudConfig",
```
brandondutra: maybe do namedtuple('name', ['a', 'b']) so you don't have to use strings (and accidentally forget a space in the future).

solutionbox/inception/datalab_solutions/inception/_cloud.py
```diff
@@ -103,10 +102,9 @@ def train(self, input_dir, batch_size, max_steps, output_path,
     job_request = {
       'package_uris': staging_package_url,
       'python_module': 'datalab_solutions.inception.task',
-      'scale_tier': scale_tier,
-      'region': region,
       'args': job_args
     }
+    job_request.update(dict(cloud_train_config._asdict()))
```
brandondutra: does the service print a clear error message if both scale tier and worker_count are set?

datalab/mlalpha/_cloud_training_config.py
qimingj: Done.

solutionbox/inception/datalab_solutions/inception/_cloud.py
qimingj: Probably yes. But I don't want to get in the business of validating the values. Service should do that.

PR #179, opened 2017-02-09
Renaming federated tables to external data sources.

chmeyers:


google/datalab/bigquery/_external_data_source.py
```diff
@@ -18,7 +18,7 @@
 from . import _csv_options
 
 
-class FederatedTable(object):
+class ExternalDataSource(object):
 
   @staticmethod
   def from_storage(source, source_format='csv', csv_options=None, ignore_unknown_values=False,
```
yebrahim: while at it, can we just make this the constructor? i'm not sure why we have a factory method if there's only one way of creating the object

chmeyers: Done.

PR #181, opened 2017-02-10
Rename 'values' to 'env', and 'query_params' in Query constructor

yebrahim:
In preparation for adding BigQuery parameterization, we'll need to maintain two dictionaries in `Query`, one for resolving references to other objects (UDFs, datasources...), and one for query parameters the user can pass.

### DIFF
--- a/google/datalab/bigquery/_query.py
+++ b/google/datalab/bigquery/_query.py
@@ -38,13 +38,7 @@ class Query(object):
     """Initializes an instance of a Query object.
 
     Args:
-      sql: the BigQuery SQL query string to execute, or a SqlStatement object. The latter will
-          have any variable references replaced before being associated with the Query (i.e.
-          once constructed the SQL associated with a Query is static).
-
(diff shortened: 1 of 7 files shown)


google/datalab/bigquery/_query.py
```diff
@@ -45,32 +45,35 @@ def __init__(self, sql, values=None, udfs=None, data_sources=None,
...
 
     if data_sources is None:
       data_sources = {}
 
     self._code = None
     self._imports = []
-    if self._values is None:
-      self._values = {}
-
-    self._sql = google.datalab.data.SqlModule.expand(sql, self._values)
```
ojarjur: It looks like the tests rely on this method being called in order to convert instances of SqlStatement to strings.

If we get rid of it, then we still need to do that conversion.

yebrahim: We're removing SqlStatement and SqlModules entirely since we're getting rid of all parsing, this will be my next PR.

PR #185, opened 2017-02-11
Support for BigQuery parameterization

yebrahim:
This PR adds support for query parameterization by passing a dictionary of parameters to BigQuery. parameters are inserted into execute calls, and look like:
```
%%bq execute --query myquery
parameters:
- name: param1
  type: STRING
  value: myvalue
- name: param2
...
```
Note the format of the schema is simpler than the one used by BigQuery, and is parsed into that format before sending to the service. The PR adds a jsonschema schema to (superficially) validate the input format before sending it to BigQuery.

Note this only adds support for named parameters. Positional parameters aren't supported yet.

### DIFF
--- a/google/datalab/bigquery/commands/_bigquery.py
+++ b/google/datalab/bigquery/commands/_bigquery.py
@@ -25,6 +25,7 @@ try:
 except ImportError:
   raise Exception('This module can only be loaded in ipython.')
 
+import jsonschema
 import fnmatch
 import json
 import re
@@ -35,6 +36,26 @@ from google.datalab.bigquery._sampling import Sampling
 import google.datalab.data
 import google.datalab.utils
 import google.datalab.utils.commands
+
+query_params_schema = {
+  'type': 'object',
+  'properties': {
+    'queryParameters': {
+      'type': 'array',
+      'items': {
+        'type': 'object',
+        'properties': {
+          'name': {'type': 'string'},
+          'parameterType': {'type': 'object'},
+          'parameterValue': {'type': 'object'}
+        },
+        'required': ['name', 'parameterType', 'parameterValue'],
+        'additionalProperties': False
+      }
+    }
+  },
+  'required': ['queryParameters']
+}
 
 
 def _create_dataset_subparser(parser):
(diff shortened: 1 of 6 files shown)


google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -155,6 +155,7 @@ def _create_query_subparser(parser):
   query_parser.add_argument('--datasources', help='List of external datasources to reference in the query body',
                             nargs='+')
   query_parser.add_argument('--subqueries', help='List of subqueries to reference in the query body', nargs='+')
+  query_parser.add_argument('--parameters', help='List of subqueries to reference in the query body', nargs='+')
```
parthea: s/List of subqueries/List of parameters/

nikhilk: I think parameters realistically need to go into the cell body.

1. They won't necessarily fit on the one line.
2. You need to support both positional as well as named parameters. Commandline won't help with latter.

yebrahim: @parthea thanks for catching that

@nikhilk We also need the query body to go in the cell. One solution I see is using yaml syntax like we do in other magics, something like:
```
%%bq query --name myquery
- query: SELECT * FROM `mydataset.mytable WHERE field1 = @param1 AND field2 >= @param2
- parameters:
  - name: param1
    type: STRING
    value: :value
  - name: param2
    type: INT64
    value: 50
```
However, I feel like putting more things in the cell body is making writing magics more difficult, it's not really much easier than using the module API anymore. It also takes away from the spirit of writing SQL freely without worrying about delimiters and line breaks, but using the above syntax will push you to think about that. Do you have a better idea?

I think using a dictionary (or list for positional?) to contain parameters is a better experience than putting everything in the cell body.

nikhilk:
I would expect parameters to be specified in bq execute. Not when defining
a query in bq query (same holds true in API).

Execute is where the binding of parameters to values happens. So in that
sense bq query is just SQL, and bq execute is just yaml syntax for
parameters.

yebrahim:
@nikhilk sounds good, I'll switch to passing parameters to execute.

yebrahim:
This is now ready for review.

nikhilk:
Suggest:

```
%bq execute  ....
parameters
- name: p1
  type: STRING
  value: some_value
```

PR #187, opened 2017-02-13
CsvDataSet no longer globs files in init.

brandondutra:


brandondutra:
ok, updated the class to not always glob the input file list

datalab/mlalpha/_dataset.py
```diff
@@ -59,21 +60,34 @@ def __init__(self, files, schema=None, schema_file=None):
...
+
+  @property
+  def _input_files(self):
+    """Returns the file list that was given to this class without globing files."""
+    return self._input_files
 
   @property
   def files(self):
-    return self._files
+    if self._glob_files:
```
qimingj: Is:

if not self._glob_files:
  self._glob_files += ...

return self._glob_files

better?

datalab/mlalpha/_dataset.py
```diff
@@ -59,21 +60,34 @@ def __init__(self, files, schema=None, schema_file=None):
...
+      return self._glob_files
+
+    for file in self._input_files:
+      # glob_files() returns unicode strings which doesn't make DataFlow happy. So str().
+      self._files += [str(x) for x in ml.util._file.glob_files(file)]
```
qimingj: Should be self._glob_files?

datalab/mlalpha/_dataset.py
brandondutra: ops, yes

brandondutra:
ok, now done :)

PR #189, opened 2017-02-13
Renaming from api review of datalab.bigquery.Query

chmeyers:
renaming execute_dry_run and scripts

### DIFF
--- a/google/datalab/bigquery/_query.py
+++ b/google/datalab/bigquery/_query.py
@@ -154,11 +154,11 @@ class Query(object):
     return self._expanded_sql()
 
   @property
-  def scripts(self):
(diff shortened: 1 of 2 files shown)


google/datalab/bigquery/_query.py
```diff
@@ -154,11 +154,11 @@ def sql(self):
     return self._expanded_sql()
 
   @property
-  def scripts(self):
+  def udfs(self):
     """ Get the code for any Javascript UDFs used in the query. """
     return self._code
```
yebrahim: this should return `self._udfs`, i forgot to make that change. we should remove `self._code` now since we don't support legacy SQL anymore.

PR #192, opened 2017-02-14
Move cloud trainer and predictor from their own classes to Job and Model respectively.

qimingj:
Cloud trainer and predictor will be cleaned up in a seperate change.

### DIFF
--- a/datalab/mlalpha/_cloud_models.py
+++ b/datalab/mlalpha/_cloud_models.py
@@ -222,7 +222,32 @@ class CloudModelVersions(object):
     name = ('%s/versions/%s' % (self._full_model_name, version_name))
     response = self._api.projects().models().versions().delete(name=name).execute()
     self._wait_for_long_running_operation(response)
-    
+
+  def predict(self, version_name, data):
+    """Get prediction results from features instances.
+
+    Args:
+      version_name: the name of the version used for prediction.
+      data: typically a list of instance to be submitted for prediction. The format of the
+          instance depends on the model. For example, structured data model may require
+          a csv line for each instance.
+    Returns:
+      A list of prediction results for given instances. Each element is a dictionary representing
+          output mapping from the graph.
+      An example:
+        [{"predictions": 1, "score": [0.00078, 0.71406, 0.28515]},
+         {"predictions": 1, "score": [0.00244, 0.99634, 0.00121]}]
+    """
+    full_version_name = ('%s/versions/%s' % (self._full_model_name, version_name))
+    request = self._api.projects().predict(body={'instances': data},
+                                           name=full_version_name)
+    request.headers['user-agent'] = 'GoogleCloudDataLab/1.0'
+    result = request.execute()
+    if 'predictions' not in result:
+      raise Exception('Invalid response from service. Cannot find "predictions" in response.')
+
+    return result['predictions']
+
   def describe(self, version_name):
     """Print information of a specified model.
 
--- a/datalab/mlalpha/_cloud_models.py
+++ b/datalab/mlalpha/_cloud_models.py
@@ -27,13 +27,11 @@ _CLOUDML_DISCOVERY_URL = 'https://storage.googleapis.com/cloud-ml/discovery/' \
                          'ml_v1beta1_discovery.json'
 
 
-class CloudModels(object):
+class Models(object):
   """Represents a list of Cloud ML models for a project."""
 
   def __init__(self, project_id=None):
-    """Initializes an instance of a CloudML Model list that is iteratable
-           ("for model in CloudModels()").
-
+    """
     Args:
       project_id: project_id of the models. If not provided, default project_id will be used.
     """
@@ -51,7 +49,9 @@ class CloudModels(object):
     page_token = list_info.get('nextPageToken', None)
     return models, page_token
 
-  def __iter__(self):
+  def get_iterator(self):
+    """Get iterator of models so it can be used as "for model in Models().get_iterator()".
+    """
     return iter(datalab.utils.Iterator(self._retrieve_models))
 
   def get_model_details(self, model_name):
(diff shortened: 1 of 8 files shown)


datalab/mlalpha/_job.py
```diff
@@ -68,6 +64,62 @@ def describe(self):
...
+        job_id = new_job_request['python_module'].replace('.', '_') + \
+            '_' + job_id
+        
+    job = {
+        'job_id': job_id,
+        'training_input': new_job_request,
+    }
+    context = datalab.context.Context.default()
+    cloudml = discovery.build('ml', 'v1beta1', credentials=context.credentials,
+                              discoveryServiceUrl=_CLOUDML_DISCOVERY_URL)
```
chmeyers: Were you going to remove discoveryServiceUrl here?

qimingj: yes. That will be a separate change (from v1beta1 to v1).

datalab/mlalpha/_job.py
```diff
@@ -68,6 +51,61 @@ def describe(self):
     job_yaml = yaml.safe_dump(self._info, default_flow_style=False)
     print job_yaml
 
+  @staticmethod 
+  def submit_training(job_request, job_id=None):
+    """Submit a training job.
```
brandondutra: expand comment that the args dict will get expanded to 
--key=value or --key=list_item_0 --key=list_item_1, ...

Also say that the python_module "trainer/task.py" should be able to parse these args.

Wait, regular flags are not supported? So task.py cannot use just "--foo". Only key-value parameters are supported? I don't like this restriction. I also bet we don't document this restriction anywhere. 

So command lines that are not supported in datalab:
1) --foo
2) --foo 1 2 3
3) --foo=--bar=2  (because we append the key and value in two calls to .append)
4) -n
5) -n 200

I would rather allow args to be a dict or list. So then we support every kind of command line. Or we should at least document task.py arg's restrictions

datalab/mlalpha/_cloud_models.py
```diff
@@ -51,7 +43,9 @@ def _retrieve_models(self, page_token, page_size):
     page_token = list_info.get('nextPageToken', None)
     return models, page_token
 
-  def __iter__(self):
+  def get_iterator(self):
```
brandondutra: why the name change to get_iterator()? Also do the same change to the Jobs class.

solutionbox/inception/datalab_solutions/inception/_cloud.py
```diff
@@ -122,8 +122,7 @@ def predict(self, model_id, image_files):
         'image_bytes': {'b64': img}
       })
 
-    cloud_predictor = mlalpha.CloudPredictor(parts[0], parts[1])
-    predictions = cloud_predictor.predict(data)
+    predictions = mlalpha.ModelVersions(parts[0]).predict(parts[1], data)
     if len(predictions) == 0:
```
brandondutra: better error message if image_files is empty/None?

datalab/mlalpha/_cloud_models.py
qimingj: Done on Jobs class. The reason of the change is that having a class both exposing methods and exposing itself as iterator is a bit odd and the iterator function is less discoverable.

datalab/mlalpha/_cloud_models.py
```diff
@@ -51,7 +43,9 @@ def _retrieve_models(self, page_token, page_size):
...
+    """Get iterator of models so it can be used as "for model in Models().get_iterator()".
+    """
     return iter(datalab.utils.Iterator(self._retrieve_models))
 
   def get_model_details(self, model_name):
```
qimingj: Added a bit in comments. create() is instant (it does not return a long running operation) so there is no need to return anything.

datalab/mlalpha/_cloud_models.py
```diff
@@ -152,7 +143,10 @@ def _retrieve_versions(self, page_token, page_size):
...
     return versions, page_token
 
-  def __iter__(self):
+  def get_iterator(self):
+    """Get iterator of versions so it can be used as
+       "for v in ModelVersions(model_name).get_iterator()".
+    """
     return iter(datalab.utils.Iterator(self._retrieve_versions))
 
   def get_version_details(self, version_name):
```
qimingj: It does. "return self._api.projects().models().versions().get(name=name).execute()"? It returns what the REST API call returns.

datalab/mlalpha/_cloud_models.py
qimingj: Now I think it is better to return whatever service returns, so it is consistent across the methods.

datalab/mlalpha/_cloud_models.py
```diff
@@ -222,7 +216,32 @@ def delete(self, version_name):
     name = ('%s/versions/%s' % (self._full_model_name, version_name))
     response = self._api.projects().models().versions().delete(name=name).execute()
     self._wait_for_long_running_operation(response)
-    
+
+  def predict(self, version_name, data):
+    """Get prediction results from features instances.
```
qimingj: Added.

datalab/mlalpha/_job.py
qimingj: Done on comments. I think we only expand args if it is a dict. If it is a list we don't mess up with it.

datalab/mlalpha/_job.py
```diff
@@ -68,6 +51,61 @@ def describe(self):
...
+              'output_path': 'gs://mubucket/data/mymodel/',
+            }
+          }
+      job_id: id for the training job. If None, an id based on timestamp will be generated.
+    Returns:
+      A Job object representing the cloud training job.
+    """
+    new_job_request = dict(job_request)
+    # convert job_args from dict to list as service required.
+    if 'args' in job_request and isinstance(job_request['args'], dict):
```
qimingj: If someone passes in a list, we pass it through to the service, and service will respond with errors if it is not expected.

datalab/mlalpha/_job.py
```diff
@@ -68,6 +51,61 @@ def describe(self):
...
+      job_args = job_request['args']
+      args = []
+      for k,v in job_args.iteritems():
+        if isinstance(v, list):
+          for item in v:
+            args.append('--' + k)
```
qimingj: k is the name of the arg so it should always be a string. I added str() to be safe.

solutionbox/inception/datalab_solutions/inception/_cloud.py
qimingj: Done.

qimingj:
PTAL. Thanks!

PR #199, opened 2017-02-14
removed the feature type file 

brandondutra:
and cloud batch prediction works now.

todo:
make transforms file optional (or pass a dict vs a file)
use datasets and cloud datasets
make cloud online prediction inspect the input to see if target is missing or not

### DIFF
--- a/solutionbox/structured_data/datalab_solutions/structured_data/preprocess/cloud_preprocess.py
+++ b/solutionbox/structured_data/datalab_solutions/structured_data/preprocess/cloud_preprocess.py
@@ -26,9 +26,7 @@ import sys
 
 from tensorflow.python.lib.io import file_io
 
-INPUT_FEATURES_FILE = 'input_features.json'
 SCHEMA_FILE = 'schema.json'
-
 NUMERICAL_ANALYSIS_FILE = 'numerical_analysis.json'
 CATEGORICAL_ANALYSIS_FILE = 'vocab_%s.csv'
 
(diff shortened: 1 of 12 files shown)


brandondutra:
removed the feature type file:

The datatype in the schema file sets numerical/categorical variables. All strings are treated like categorical columns, and float/int are numeric. Preprocessing does not really know which target is target or key.

At preprocessing time, the transforms file lists the target column (which is a little redundant) and the key column. These currently have to be there. Defaults are also set in the transformed file, but are optional.

Example transforms file to dnn classification:
    transforms = {
        "num1": {"transform": "scale", "default": 0.0},
        "num2": {"transform": "scale","value": 4},
        "str1": {"transform": "hash_one_hot", "hash_bucket_size": 4},
        "str2": {"transform": "one_hot", "default": "blue"},
        "str3": {"transform": "embedding", "embedding_dim": 3},
        "target": {"transform": "target"},
        "key": {"transform": "key"}
    }

Defaults for categorical columns are the empty string/label, and defaults for numerical columns is that column's mean (from preprocessing)

brandondutra:
My training packing building is a little different, I just copy setup.py to be in the installed location. So I no longer assume this is running in the VM (I think :) )

solutionbox/structured_data/datalab_solutions/structured_data/preprocess/cloud_preprocess.py
```diff
@@ -169,16 +168,19 @@ def run_categorical_analysis(table, args, feature_types):
   ...
 
   Args:
-    table: Reference to FederatedTable if bigquery_table is false.
+    table: Reference to FederatedTable if bigquery_table is false
```
qimingj: What does "if bigquery_table is false" mean?

brandondutra: updated comment.

PR #202, opened 2017-02-14
List all objects in storage when using gs:// prefix

yebrahim:
This makes the `%gcs list` command behave more like the `gsutil` tool in that it can list GCS paths with a given prefix if that prefix is a valid path. This does not change its flexible capabilities, so it can still use wild characters to list objects.

### DIFF
--- a/google/datalab/storage/commands/_storage.py
+++ b/google/datalab/storage/commands/_storage.py
@@ -95,12 +95,8 @@ for help on a specific command.
 
   list_parser = parser.subcommand('list', 'List buckets in a project, or contents of a bucket.')
   list_parser.add_argument('-p', '--project', help='The project associated with the objects')
-  group = list_parser.add_mutually_exclusive_group()
-  group.add_argument('-o', '--object',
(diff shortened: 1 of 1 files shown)


google/datalab/storage/commands/_storage.py
```diff
@@ -95,12 +95,8 @@ def gcs(line, cell=None):
...
   list_parser = parser.subcommand('list', 'List buckets in a project, or contents of a bucket.')
   list_parser.add_argument('-p', '--project', help='The project associated with the objects')
-  group = list_parser.add_mutually_exclusive_group()
-  group.add_argument('-o', '--object',
-                     help='The name of the objects(s) to list; can include wildchars',
-                     nargs='?')
-  group.add_argument('-b', '--bucket',
-                     help='The name of the buckets(s) to list; can include wildchars',
+  list_parser.add_argument('-o', '--objects',
+                     help='List objects under the given Google Cloud Storage path',
```
chmeyers: Indent to match open parens?

PR #203, opened 2017-02-15
Import fixes from master

yebrahim:
These are some fixes done in master, this PR is importing them and applying them for both namespaces `datalab` and `google.datalab`.


yebrahim:
I changed this to includ only Bradley's change in master to drop TF and CloudML SDK from requirements because it improves the pip install and therefore tests.

PR #208, opened 2017-02-16
Use workaround a TF summary issue. Force online prediction to use TF 1.0

qimingj:


qimingj:
Will send another one. This one seems corrupted.

PR #210, opened 2017-02-16
Fix chart object lookup

yebrahim:


ojarjur:
I'm not sure what this is trying to fix; can you add a link to an issue?

google/datalab/utils/commands/_chart.py
```diff
@@ -52,6 +53,9 @@ def chart(line, cell=None):
 
 def _chart_cell(args, cell):
   source = args['data']
+  item = google.datalab.utils.commands.get_notebook_item(source)
```
ojarjur: One of the current use cases is charting literal data, and this change seems to break that.

Can you wrap this in a check to see if the type is a string before trying to look it up from the notebook environment?

yebrahim: You're right, I didn't notice that. Will drop this PR.

PR #215, opened 2017-02-17
Add a resize option for inception package to avoid sending large data to online prediction

qimingj:
Also update Lantern browser.


solutionbox/inception/datalab_solutions/inception/_package.py
```diff
@@ -142,44 +142,49 @@ def cloud_train(input_dir, batch_size, max_steps, output_dir,
...
 
 
-def local_predict(model_dir, image_files, show_image=True):
+def local_predict(model_dir, image_files, resize=False, show_image=True):
   """Predict using an offline model.
   Args:
     model_dir: The directory of a trained inception model. Can be local or GCS paths.
     image_files: The paths to the image files to predict labels. Can be local or GCS paths.
     show_image: Whether to show images in the results.
+    resize: Whether to resize the image to a reasonable size before prediction.
```
brandondutra: what is reasonable, can I set set the size. What happens if the image is smaller than expected? Just add these points to the comment.

solutionbox/inception/datalab_solutions/inception/_util.py
```diff
@@ -184,3 +186,54 @@ def get_sources_from_dataset(p, dataset, mode):
...
   else:
     raise ValueError('Invalid DataSet. Expect CsvDataSet or BigQueryDataSet')
+
+
+def decode_and_resize(image_str_tensor):
+  """Decodes jpeg string, resizes it and returns a uint8 tensor."""
+
+  # These constants are set by Inception v3's expectations.
+  height = 299
+  width = 299
```
brandondutra: not 300? is there some kind of "start counting at zero" going on here? Add comment if so.

solutionbox/inception/datalab_solutions/inception/_util.py
```diff
@@ -184,3 +186,54 @@ def get_sources_from_dataset(p, dataset, mode):
...
+  image = tf.squeeze(image, squeeze_dims=[0])
+  image = tf.cast(image, dtype=tf.uint8)
+  return image
+
+
+def resize_image(image_str_tensor):
+  """Decodes jpeg string, resizes it and re-encode it to jpeg."""
+
+  image = decode_and_resize(image_str_tensor)
+  image = tf.image.encode_jpeg(image, quality=100)
```
brandondutra: could decode_and_resize do
1) decode_jpeg
2) stuff
3) encode_jpeg

so then decode_and_resize can be renamed resize_image

solutionbox/inception/datalab_solutions/inception/_package.py
qimingj: Done. TF takes care of the resizing so I believe if image is smaller it will be resized correctly anyway.

solutionbox/inception/datalab_solutions/inception/_util.py
qimingj: It's just Inception expects such dimension. Don't know the reason behind but there is already comments saying this is expected by Inception.

solutionbox/inception/datalab_solutions/inception/_util.py
qimingj: decode_and_resize() is also used by model (it does not need to encode back) so this function can only do 1 and 2.

PR #216, opened 2017-02-21
Add %bq datasource

yebrahim:
- Added `%bq datasource` magic that takes an external datasource schema
- `%bq query` now takes `--datasources`, which accepts at most one datasource for now, since the BigQuery API only allows only temporary datasource in the query insert API.
- Cleaned up `__str__` definitions and only kept `__repr__`, which is used by Jupyter to show string representations of objects. Calling `str()` on an object will end up calling its `__repr__`.
- When creating a Query object, references to subqueries, UDFs, and datasource are now validated by instance type as well.
- Fixed tests to use `_repr_sql_()` when printing table/view names, since they shouldn't care about our decorated string values of the objects.

### DIFF
--- a/google/datalab/bigquery/_query.py
+++ b/google/datalab/bigquery/_query.py
@@ -25,6 +25,7 @@ from . import _query_job
 from . import _sampling
 from . import _udf
 from . import _utils
+from . import _external_data_source
 
 
 class Query(object):
@@ -33,8 +34,7 @@ class Query(object):
   This object can be used to execute SQL queries and retrieve results.
   """
 
-  def __init__(self, sql, env=None, udfs=None, data_sources=None,
-               subqueries=None):
+  def __init__(self, sql, env=None, udfs=None, data_sources=None, subqueries=None):
     """Initializes an instance of a Query object.
 
     Args:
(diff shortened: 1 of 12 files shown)


google/datalab/bigquery/_query.py
```diff
@@ -33,49 +34,47 @@ class Query(object):
...
 
     Raises:
       Exception if expansion of any variables failed.
       """
     self._sql = sql
-    self._data_sources = data_sources
     self._udfs = udfs
     self._subqueries = subqueries
     self._env = env or {}
+    self._data_sources = []
```
ojarjur: Here we are setting `self._data_sources` to be a list, but below it is set to a dictionary.

We should be consistent and set it to an empty dictionary here (or None).

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -190,6 +190,19 @@ def _create_udf_subparser(parser):
   return udf_parser
 
 
+def _create_datasource_subparser(parser):
+  datasource_parser = parser.subcommand('datasource', 'Create a named Javascript BigQuery external data source')
+  datasource_parser.add_argument('-n', '--name', help='The name for this data source',
+                                 required=True)
+  datasource_parser.add_argument('-p', '--paths', help='URL(s) of the data objects, can include ' + \
+                                 'a whilecard "*" at the end', required=True, nargs='+')
```
ojarjur: s/whilecard/wildcard/

yebrahim: Done

google/datalab/bigquery/_query.py
yebrahim: Thanks for catching this. Done.

PR #217, opened 2017-02-21
cast string to int

benopendoor:
`table.insert_data(df)` inserts data correctly but raises TypeError: unorderable types: str() > int()


benopendoor:
I signed it

PR #223, opened 2017-02-22
Remove project_id from the init methods in Stackdriver related classes.

supriyagarg:


google/datalab/stackdriver/commands/_monitoring.py
```diff
@@ -106,3 +107,10 @@ def _render_dataframe(dataframe):
   fields = dataframe.columns.tolist()
   return IPython.core.display.HTML(
       google.datalab.utils.commands.HtmlBuilder.render_table(data, fields))
+
+
+def _make_context(project_id):
+  context = google.datalab.Context.default()
```
yebrahim: This way you're setting the project id for the global context object, which will affect all other operations. Instead, what you want is copy the default context object into a new one, and change that one's project and use it.

yebrahim:
LGTM except for that one place where the global context object is modified.

google/datalab/stackdriver/commands/_monitoring.py
supriyagarg: Fixed, and merged the changes. PTAL, and merge it this is fine.

PR #225, opened 2017-02-23
Remove CloudML SDK from Dependency of Inception

qimingj:


solutionbox/inception/datalab_solutions/inception/_util.py
```diff
@@ -42,12 +43,41 @@ def default_project():
   return context.project_id
 
 
+def open_local_or_gcs(path, mode):
```
brandondutra: can you use TF file_io for this too? Less code to test later

qimingj: yes. Let's try that in structured_data first and if it works out, we don't need to depend on DataFlow in training job.

PR #228, opened 2017-02-23
TensorFlow's save_model no longer creates export.meta, so disable the  check in deploying models.

qimingj:


datalab/mlalpha/_cloud_models.py
```diff
@@ -180,11 +180,12 @@ def deploy(self, version_name, path):
     """
     if not path.startswith('gs://'):
       raise Exception('Invalid path. Only Google Cloud Storage path (gs://...) is accepted.')
-    if not datalab.storage.Item.from_url(os.path.join(path, 'export.meta')).exists():
-      # try appending '/model' sub dir.
+
+    # If there is no "export.meta" under path but there is path/model/export.meta, then
+    # append /model to the path.
+    if (not datalab.storage.Item.from_url(os.path.join(path, 'export.meta')).exists() and
+        datalab.storage.Item.from_url(os.path.join(path, 'model', 'export.meta')).exists()):
```
brandondutra: not correct. Here is my folder:

ls tout/model/
assets.extra  saved_model.pb  variables

So if I pass tout/, this does not work (as model is not added). If i pass tout/model this works as the path is not changed.

qimingj: I will check saved_model.pb too then.

qimingj:
PTAL

PR #234, opened 2017-02-24
Expose Table.to_file and Table.to_dataframe

yebrahim:
I'm not sure why we made these internal, they should be part of the public API.


ojarjur:
The test failure indicates that there are still some references to "_to_dataframe"

PR #244, opened 2017-02-27
updated batch size issue for structured data

brandondutra:
With

 tf.contrib.learn.Experiment(eval_steps=None) and 

input_reader_for_eval = get_reader_input_fn(num_epochs=1),

one pass over the eval data is done. Updated default to be small and added comment that the last fractional batch is not used. This is a fun fact of using tf.learn.


brandondutra:
not to self: todo: if batch size is larger than dataset size, fail.

brandondutra:
updated to 16

PR #254, opened 2017-03-02
mltoolbox namespace change.

brandondutra:


brandondutra:
build
1) python setup.py sdist from pydatalab/solutionbox/structured_data 
2) python setup.py sdist from pydatalab/solutionbox/image_classification
3) make a new virtual env. Pip install both tar balls
4) start python and do 
import mltoolbox.datalab_image_classification  as x  
import mltoolbox._structured_data as y

PR #255, opened 2017-03-02
Job class updates

qimingj:
Add a DataFlowJob class so we can return it from package's methods which use cloud dataflow.

Modify ml's Job class so it is based on utils.Job.

### DIFF
--- a/datalab/utils/_dataflow_job.py
+++ b/datalab/utils/_dataflow_job.py
@@ -34,8 +34,8 @@ class DataflowJob(_job.Job):
 
     # DataFlow's DataflowPipelineResult does not refresh state, so we have to do it ourselves
     # as a workaround.
-    self._runner_results._job = \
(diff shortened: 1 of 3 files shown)


datalab/utils/_dataflow_job.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    super(DataflowJob, self).__init__(runner_results._job.name)
+    self._runner_results = runner_results
+
+  def _refresh_state(self):
+    """ Refresh the job info. """
+
+    # DataFlow's DataflowPipelineResult does not refresh state, so we have to do it ourselves
+    # as a workaround.
+    self._runner_results._job = \
```
brandondutra: instead of \, you could also do 

..._job = (
    stuff
)

This prevents accidentally needing more than one line.

qimingj: done.

PR #256, opened 2017-03-02
Update image classification package.

qimingj:
- Namespace change.
- Use a cloud config to indicate whether it is cloud run or local run.


qimingj:
This change already includes the cloud parameter. See the added _api.py. Local and Cloud are no longer exported.

PR #263, opened 2017-03-03
provide a get_list() method to make it easier to programmatically get…

lakshmanok:
… running processes


qimingj:
I had a similar change.

PR #265, opened 2017-03-03
Update Context.default() to always get latest config, credentials, and project.

qimingj:


google/datalab/_context.py
```diff
@@ -112,9 +112,15 @@ def default():
     Returns:
       An initialized and shared instance of a Context object.
     """
-    if Context._global_context is None:
-      project = get_default_project_id()
-      credentials = get_credentials()
-      config = Context._get_default_config()
+    credentials = get_credentials()
+    project = get_default_project_id()
+    config = Context._get_default_config()
```
ojarjur: This will wipe out any modifications a user has made to the default config with a static dictionary.

I don't think that is what we want.

I would just move the config setting back inside of the then branch for the subsequent if statement.

ojarjur:
I think this is addressing a real problem that we need to solve, but I'm not sure if this is the right solution.

Separately, I was looking at a related issue where the `datalab_project_id` and `set_datalab_project_id` methods (which were moved from [here](https://github.com/googledatalab/pydatalab/blob/master/datalab/kernel/__init__.py#L130) to [here](https://github.com/googledatalab/pydatalab/blob/master/google/datalab/kernel/__init__.py#L131)) were no longer including the datalab.context.Context.default() object, which may break some old notebooks.

I think both issues are probably related, and we may want to try to fix them in a single change.

Maybe consider the following:

1. Have the `project_id` method read from the `get_default_project_id` method if it is the default context (similarly to how we set that for the default context).
2. Include similar logic for the credentials (for the default context).
3. Make the old context class' default read and write to the new context class' default, if the new context class is importable.

Does that sound reasonable?

ojarjur:
Actually, I take back what I said earlier.

Now, I'm thinking that the two issues are separate if we treat the old Context class as something we are not changing (i.e. maintain backwards compatibility even if the old behavior is arguably wrong).

However, if we take that approach, then we should not be changing the old class.

In particular, I worry that changing it will break notebooks that do something like

```python
datalab.context.Context.default().set_project_id('something')
...
datalab.context.Context.default().project_id
```

PR #266, opened 2017-03-03
Fix backwards-compatibility issue with the new `Context` class.

ojarjur:
This change fixes a backwards-incompatibility that was recently
introduced regarding the `datalab_project_id`, `datalab_bq_dialect`,
`set_datalab_project_id`, and `set_datalab_bq_dialect` helper methods,
and the old `datalab.context.Context` and `datalab.bigquery.Dialect`
classes.

Previously, those helper methods set and get the default instance
of the corresponding classes. However, with the new `google.*` modules,
those methods instead get and set the default instance of the
`google.datalab.Context` class.

That would be a breaking change for existing notebooks, so we
have to make sure that those helper methods *also* set and get the
older classes.

For example, if your project ID is initially 'a', then the
following code cell:

```python
from google.datalab import Context as ctx
from datalab.context import Context as old_ctx

default_ctx = ctx.default()
default_old_ctx = old_ctx.default()

print datalab_project_id()
print default_ctx.project_id
print default_old_ctx.project_id

set_datalab_project_id('b')

print datalab_project_id()
print default_ctx.project_id
print default_old_ctx.project_id
```

... will now output:

```
a
a
a
b
b
b
```

... whereas, without the fix, it would have output:

```
a
a
a
b
b
a
```

Similarly, the following cell:

```python
default_old_ctx.set_project_id('a')
print datalab_project_id()
```

... will output `a`, as expected.

There is one strange behavior that this introduces:

Calling `datalab.context.Context.default().set_project_id` will
also change the project ID in the `google.datalab.Context.default()`
object. However, calling `google.datalab.Context.default().set_project_id`
will *not* update the project ID in the `datalab.context.Context.default()`
object.

That design has been adopted because the behavior of the old class
was rather static (the stored values only changed if explicitly changed),
but the behavior of the new class is more dynamic (e.g. the project
ID is changed if you run `gcloud config set project ...`).

Since we did not want to change the static nature of the old class,
there has to be some place where the connection between the two
versions of Context is broken, and that is the place picked to do that.

### DIFF
--- a/datalab/bigquery/_dialect.py
+++ b/datalab/bigquery/_dialect.py
@@ -36,6 +36,12 @@ class Dialect(object):
     """ Set the default BigQuery SQL dialect"""
     if bq_dialect in ['legacy', 'standard']:
       self._global_dialect = bq_dialect
+      try:
+        from google.datalab import Context as new_context
+        new_context.default().config['bigquery_dialect'] = bq_dialect
+      except ImportError:
+        # If the new library is not loaded, then we have nothing to do.
+        pass
 
   @staticmethod
   def default():
--- a/datalab/bigquery/_dialect.py
+++ b/datalab/bigquery/_dialect.py
@@ -36,12 +36,6 @@ class Dialect(object):
     """ Set the default BigQuery SQL dialect"""
     if bq_dialect in ['legacy', 'standard']:
       self._global_dialect = bq_dialect
-      try:
-        from google.datalab import Context as new_context
-        new_context.default().config['bigquery_dialect'] = bq_dialect
-      except ImportError:
-        # If the new library is not loaded, then we have nothing to do.
-        pass
 
   @staticmethod
   def default():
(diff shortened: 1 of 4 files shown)


datalab/bigquery/_dialect.py
```diff
@@ -36,6 +36,12 @@ def set_bq_dialect(self, bq_dialect):
     """ Set the default BigQuery SQL dialect"""
     if bq_dialect in ['legacy', 'standard']:
       self._global_dialect = bq_dialect
+      try:
+        from google.datalab import Context as new_context
+        new_context.default().config['bigquery_dialect'] = bq_dialect
```
yebrahim: we should remove `bigquery_dialect` everywhere, we don't support legacy SQL anymore in the new namespace.

ojarjur: OK, but the full change to remove all (the other) references to `bigquery_dialect` is very large and should be done in a separate pull request.

PR #272, opened 2017-03-05
Add syntax coloring for `%bq query` magic

yebrahim:
(This PR is a functional dependency for https://github.com/googledatalab/datalab/pull/1242)

Adds correct syntax for BigQuery Legacy SQL into a new CodeMirror mode called 'bigquery', and fixes typescript references so that `google.datalab.notebook` can be installed as a separate Jupyter extension, containing this new CodeMirror mode.

Also fixes the regex used to detect operators in the 'sql' CodeMirror mode.


google/datalab/notebook/static/codemirror/mode/bigquery.js
```diff
@@ -13,290 +13,109 @@
...
-      } else if (ch === '<' && !stream.match(/^[\s\u00a0=]/, false)) {
-        stream.match(/^[^\s\u00a0>]*>?/);
-        return 'atom';
       } else if (ch === '"' || ch === '\'') {
         state.tokenize = tokenLiteral(ch);
         return state.tokenize(stream, state);
       } else if (ch === '`') {
-        state.tokenize = tokenOpLiteral(ch);
-        return state.tokenize(stream, state);
+        stream.eatWhile(/[^`]/);
```
nikhilk: Curious why this string handling is different than the if block above for double and single quotes? And how if needed, this handles escapes?

yebrahim: Thanks for catching that. We could use the same tokenizer for all three types of quotes and treat them all as strings, so I did that.
And yes, that tokenizer handles escaping.

yebrahim:
Those changes are necessary for correctness though, and I don't feel they pose much of a risk anyway since they're about syntax coloring. Here's a breakdown of the changes to help the review:

- List of operator characters was changed from legacy to standard, and the regex used for legacy wasn't correct anyway
- Comments can start anywhere, not just beginning of the line
- `--` used to start a comment, just moved this up to group it with the '#' tokenizing
- Couldn't find an atom definition in the lexical structure that looks like `<word>`, removed this
- Added backticks to the list of string delimiters
- Couldn't find an atom definition that looks like `:word`
- Operators can also be words, added tokenizer for a list of those

PR #273, opened 2017-03-05
made x and x_async for sd package

brandondutra:
both return job object.
Notebooks need updating to not call wait(), as job is not returned. Notebooks are currently broken.

also added Nikhil's comments in his Sun, Mar 5, 2017 at 4:54 PM email.

### DIFF
--- a/solutionbox/structured_data/mltoolbox/_structured_data/_package.py
+++ b/solutionbox/structured_data/mltoolbox/_structured_data/_package.py
@@ -133,7 +133,7 @@ def analyze(output_dir, dataset, cloud=False, project_id=None):
       cloud=cloud,
       project_id=project_id)
   job.wait()
-  return job
+
 
 def analyze_async(output_dir, dataset, cloud=False, project_id=None):
   """Analyze data locally or in the cloud with BigQuery.
@@ -700,7 +700,7 @@ def batch_predict(training_output_dir, prediction_input_file, output_dir,
       output_format=output_format,
       cloud=cloud)
   job.wait()
-  return job
+
 
 def batch_predict_async(training_output_dir, prediction_input_file, output_dir,
                   mode, batch_size=16, shard_files=True, output_format='csv',
--- a/solutionbox/structured_data/mltoolbox/_structured_data/_package.py
+++ b/solutionbox/structured_data/mltoolbox/_structured_data/_package.py
@@ -89,7 +89,7 @@ def _package_to_staging(staging_package_url):
         os.path.join(os.path.dirname(__file__), '../../'))
     setup_path = os.path.abspath(
         os.path.join(os.path.dirname(__file__), 'master_setup.py'))
-    tar_gz_path = os.path.join(staging_package_url, 'staging', 'sd.tar.gz')
+    tar_gz_path = os.path.join(staging_package_url, 'staging', 'trainer.tar.gz')
 
     print('Building package and uploading to %s' % tar_gz_path)
     ml.package_and_copy(package_root, setup_path, tar_gz_path)
@@ -213,6 +213,7 @@ def train_async(train_dataset,
           learning_rate=0.01,
           epsilon=0.0005,
           job_name=None, # cloud param
+          job_name_prefix='', # cloud param
           cloud=None, # cloud param
           ):
   # NOTE: if you make a chane go this doc string, you MUST COPY it 4 TIMES in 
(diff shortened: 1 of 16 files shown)


solutionbox/structured_data/mltoolbox/_structured_data/_package.py
```diff
@@ -125,8 +125,17 @@ def _wait_and_kill(pid_to_wait, pids_to_kill):
...
-
 def analyze(output_dir, dataset, cloud=False, project_id=None):
+  """Blocking version of analyze_async. See documentation of analyze_async."""
+  job = analyze_async(
+      output_dir=output_dir,
+      dataset=dataset,
+      cloud=cloud,
+      project_id=project_id)
+  job.wait()
+  return job
```
qimingj: We don't need to return anything here. The job object is not useful once the job is completed.

solutionbox/structured_data/mltoolbox/_structured_data/_package.py
```diff
@@ -493,7 +507,8 @@ def cloud_train(train_dataset,
   job_request.update(dict(config._asdict()))
 
   if not job_name:
-    job_name = 'structured_data_train_' + datetime.datetime.now().strftime('%y%m%d_%H%M%S')
+    job_name = job_name_prefix or 'structured_data_train'
```
nikhilk: Does this ever get to not having a job name, or do the public APIs always produce an appropriately named job_name? If yes, maybe mltoolbox_train instead?

solutionbox/structured_data/mltoolbox/_structured_data/trainer/task.py
```diff
@@ -152,9 +152,8 @@ def parse_arguments(argv):
                       type=str,
                       required=True,
                       help=('Output folder of preprocessing. Should contain the'
-                            ' files input_features.json, schema.json, and the'
-                            ' optional files numerical_analysis.json and'
-                            ' vocab_str1.csv. Path must be on GCS if running'
+                            ' schema file, and numerical stats and vocab files.'
+                            ' Path must be on GCS if running'
                             ' cloud training.'))
   parser.add_argument('--transforms-file',
```
nikhilk: features-file?

solutionbox/structured_data/mltoolbox/classification/dnn/_classification_dnn.py
```diff
@@ -16,6 +17,45 @@ def train(train_dataset,
           job_name=None, 
           cloud=None, 
           ):
+  """Blocking version of train_async. See documentation for train_async."""
+  job = train_async(
+      train_dataset=train_dataset,
+      eval_dataset=eval_dataset,
+      analysis_output_dir=analysis_output_dir,
```
nikhilk: Can you call this just analysis_dir? The current name makes it sound like this is an output from train, rather than an input to train.

solutionbox/structured_data/mltoolbox/_structured_data/_package.py
```diff
@@ -673,6 +686,25 @@ def cloud_predict(model_name, model_version, data):
 def batch_predict(training_output_dir, prediction_input_file, output_dir,
                   mode, batch_size=16, shard_files=True, output_format='csv',
                   cloud=False):
+  """Blocking versoin of batch_predict. 
+
+  See documentation of batch_prediction_async.
+  """
+  job = batch_predict_async(
+      training_output_dir=training_output_dir,
```
nikhilk: Similarly, training_dir.

solutionbox/structured_data/mltoolbox/_structured_data/_package.py
brandondutra: no. When using the regression/classification package, the job_name_prefix will be something like 'mltoolbox_regression_dnn_<timestamp>"

solutionbox/structured_data/mltoolbox/_structured_data/trainer/task.py
brandondutra: yea, everything in the trainer calls it transform. Users don't call the task.py file, but they will see --transforms-file if they look at the jobs page in the web UI. I figure when I stop using tf.learn, I'll have to rewrite task.py anyway, so I'll deal with it then.

solutionbox/structured_data/mltoolbox/classification/dnn/_classification_dnn.py
brandondutra: ok, will do this in a new PR

solutionbox/structured_data/mltoolbox/_structured_data/_package.py
brandondutra: ok, will do this in a new PR

PR #280, opened 2017-03-06
Remove tensorflow, dataflow from setup.py of image package. Add them as extra package when submitting jobs.

qimingj:


brandondutra:
how is this different from just listing them in setup.py? I thought the point was to not mess with the the installed packages when training? For prediction/dataflow, yes we need to install them.

PR #289, opened 2017-03-07
confusion matrix can glob files

brandondutra:


brandondutra:
how do build a container with the change? When I run the normal

./build.sh <pydatalab>
./run

I don't get the new CM code.

qimingj:
That requires rebuilding of ga image from jenkins. I can do that once it is checked in.

PR #195, opened 2017-02-14
Use http Keep-Alive, else BigQuery queries are ~seconds slower than necessary

jdanbrown:
- Before (without Keep-Alive): ~3-7s for BigQuery `select 3` with an already cached result
- After (with Keep-Alive): ~1.5-3s
- Query sends these 6 http requests and runtime appears to be dominated by network RTT:
```
http: method[POST], url[https://www.googleapis.com/bigquery/v2/projects/foo-bar/jobs/], body[{"kind": "bigquery#job", "configuration": {"priority": "INTERACTIVE", "query": {"allowLargeResults": false, "useLegacySql": false, "useQueryCache": true, "query": "select 3", "userDefinedFunctionResources": []}, "dryRun": false}}]
http: method[GET], url[https://www.googleapis.com/bigquery/v2/projects/foo-bar/queries/job_sIi1HRqfyHkRB5_MkfWFjtZ60XM?startIndex=0&timeoutMs=30000&maxResults=0], body[None]
http: method[GET], url[https://www.googleapis.com/bigquery/v2/projects/foo-bar/jobs/job_sIi1HRqfyHkRB5_MkfWFjtZ60XM], body[None]
http: method[GET], url[https://www.googleapis.com/bigquery/v2/projects/foo-bar/datasets/_2f96775300d8858559d2bd23c05bad0392345e30/tables/anon921947a4e6645dc2b34411c365f9a45e0895d5a4], body[None]
http: method[GET], url[https://www.googleapis.com/bigquery/v2/projects/foo-bar/datasets/_2f96775300d8858559d2bd23c05bad0392345e30/tables/anon921947a4e6645dc2b34411c365f9a45e0895d5a4], body[None]
http: method[GET], url[https://www.googleapis.com/bigquery/v2/projects/foo-bar/datasets/_2f96775300d8858559d2bd23c05bad0392345e30/tables/anon921947a4e6645dc2b34411c365f9a45e0895d5a4/data?maxResults=25], body[None]
```

PR #296, opened 2017-03-08
Fix a bug where an unbounded number of credentials wrappers got added.

ojarjur:
This fixes the issue where the fact that we were sharing a single
`httplib2.Http` instance caused each request to add an additional
credentials wrapper around that single instance.

That, in turn, caused a `RuntimeError: maximum recursion depth ...`
error message once too many API calls had been made.

This fixes https://github.com/googledatalab/datalab/issues/1251

### DIFF
--- a/datalab/utils/_http.py
+++ b/datalab/utils/_http.py
@@ -19,6 +19,7 @@ from builtins import str
 from past.builtins import basestring
 from builtins import object
 
+import copy
 import datetime
 import json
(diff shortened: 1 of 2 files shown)


yebrahim:
It might be cleaner to revert https://github.com/googledatalab/pydatalab/pull/195 instead of using `copy`. LGTM otherwise.

ojarjur:
@parthea That's a fair point.

I added the fix to the older library too.

PR #304, opened 2017-03-10
Fix #301: %bq query table view is broke.

### LINKED ISSUE #301: %%bq query table view is broken
For example, run:

%bq query
SELECT * from `MyTable`

The first page is fine, but if you click "next page" button, it fails with "No Data" error message.

qimingj:
Fix #301. There are two issues:

- Like old library, the new library still uses "str(table)" to get table full name, but __str__() method is removed from new library's Table class.
- Old library and new library shares the same in-memory object for storing opened tables, but due to different naming convention between old table and new table full name (mainly project:dataset.table vs project.dataset.table), it no longer works. Either old or new library breaks without separating the two.

The fix takes care of both issues above to make both "%%sql" and "%%bq query" working --- the output is a paged table that can go next and previous.

### DIFF
--- a/datalab/utils/commands/_chart_data.py
+++ b/datalab/utils/commands/_chart_data.py
@@ -30,7 +30,9 @@ import datalab.utils
 from . import _utils
 
 
-@IPython.core.magic.register_cell_magic
+# Disable the magic here because another one with same name is available under
+# google.datalab namespace.
+# @IPython.core.magic.register_cell_magic
 def _get_chart_data(line, cell_body=''):
 
   refresh = 0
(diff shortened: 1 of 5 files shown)


datalab/utils/commands/_chart_data.py
```diff
@@ -30,7 +30,9 @@
 from . import _utils
 
 
-@IPython.core.magic.register_cell_magic
+# Disable the magic here because another one with same name is available under
```
yebrahim: why not delete the whole thing?

qimingj: I want to keep it here to so the changes for datalab.bigquery to work independently are minimal.

PR #309, opened 2017-03-14
added sd training tests to pydatalab test suite

brandondutra:
This only tests the core trainer locally (local_preprocess.py and task.py) not the datalab interfaces.

All 4 training cases (linear+regression, dnn+regression, ...) are tested on generated data with numerical and categorical columns. 

Note that these tests are only done in the python 2.7 test because the structure data package has some larger issues with python 3 that should be fixed in a new PR. (for example,   json.loads(file_io.read_file_to_string(args.schema_file)) errors saying TypeError: the JSON object must be str, not 'bytes')

There are a lot of commits in this PR because I had issues with installing tensorflow in the travis build. The solution was to fix the distribution.

To do in future PRs:
--add mocked calls to the datalab interface functions (like checking for errors)
--add a notebook e2e test (in the notebook repro)
--make structured data work in python 3


solutionbox/structured_data/test_mltoolbox/test_sd_trainer.py
```diff
@@ -29,6 +32,17 @@ class TestTrainer(unittest.TestCase):
...
   """
+  def __init__(self, *args, **kwargs):
+    super(TestTrainer, self).__init__(*args, **kwargs)
+
+    # Allow this class to be subclassed for quick tests that only care about
+    # training working, not model loss/accuracy.
+    self._max_steps = 2500
+    self._check_model_loss = True
+
+    # Print test output. Set to false for a silent test.
```
yebrahim: Is this comment correct?

brandondutra: fixed

tests/mltoolbox_structured_data/traininglib_tests.py
```diff
@@ -0,0 +1,42 @@
...
+
+  def __init__(self, *args, **kwargs):
+    super(TestCoreTrainingLib, self).__init__(*args, **kwargs)
+
+    # Test that training works, not that the model is good.
+    self._max_steps = 50  
+    self._check_model_loss = False
+
+    # Don't print anything. Set to false for debugging.
+    self._silent_output = True
```
qimingj: Why do we want to suppress the debugging traces in tests?

solutionbox/structured_data/test_mltoolbox/test_sd_trainer.py
```diff
@@ -101,21 +115,27 @@ def _check_training_screen_output(self, accuracy=None, loss=None):
...
         last_line = line
         break
+
     if not last_line:
       print('Skipping _check_training_screen_output as could not find last eval'
             ' line')
       return
     print(last_line)
 
+    if not self._check_model_loss:
```
qimingj: Move this to the beginning of the function? It seems if self._check_model_loss is False, we don't check anything in this function? Or, should we move the check outside the function, so that:

if self._check_model_loss:
  self._check_training_screen_output()?

qimingj: _check_model_loss seems misleading since you also check something else if enabled, such as global step. Also the function takes "loss" argument, and if it is None it does not check loss. So there is already a flag for that?

tests/mltoolbox_structured_data/traininglib_tests.py
qimingj: Why do we want to suppress debugging info in tests?

.travis.yml
```diff
@@ -1,15 +1,27 @@
...
+
 language: python
+
 python:
   - 2.7
   - 3.5
 
 before_install:
   - sudo apt-get install -y python-setuptools
+  - strings /usr/lib/x86_64-linux-gnu/libstdc++.so.6 | grep GLIBCXX
```
yebrahim: i don't think this is needed, we can just take it out. same for the `pip freeze` command

brandondutra: done

solutionbox/structured_data/test_mltoolbox/e2e_functions.py
```diff
@@ -134,7 +134,15 @@ def make_preprocess_schema(filename, problem_type):
     f.write(json.dumps(schema))
 
 
-def run_preprocess(output_dir, csv_filename, schema_filename):
+def run_preprocess(output_dir, csv_filename, schema_filename, logger):
+  """Run preprocess via subprocess call to local_preprocess.py
+
+  Args:
+    output_dir: folder to write output to
+    csv_filename: file to do analysis on
```
qimingj: Specify whether GCS path is supported? Same for schema_filename.

brandondutra:
moved to use python's logging. Logging is set to debug in the test code, and test/main.py overrides this to info level. So this test [int test/main.py] still prints nothing, but exceptions will surface. If ran via 'python -m test_mltoolbox.test_sd_trainer', debug info is printed, which is useful at development time.

PTAL

solutionbox/structured_data/test_mltoolbox/e2e_functions.py
brandondutra: done for every parameter in this file.

PR #313, opened 2017-03-17
Remove unused imports. Fix F401 flake8 warnings.

parthea:
Fixed all flake8 warnings with code F401 (module imported but unused) that appear with the following command:

```
flake8 --select=F401
```

- Remove unused imports
- In `__init__.py`, add list of module names that should be imported when `from package import *` is encountered ( From  https://docs.python.org/2/tutorial/modules.html#importing-from-a-package ) . There is additional discussion about this approach here: https://bugs.launchpad.net/pyflakes/+bug/1178905


yebrahim:
This is great. Cleanup is always much appreciated.

While at it, can we also add the flake command to the travis file so we don't have to batch cleanup again?

parthea:
> While at it, can we also add the flake command to the travis file so we don't have to batch cleanup again?

Done for code F401 (unused imports) . I'll work on cleaning up the other flake8 codes so we can just run `flake8 .` without any filters.

yebrahim:
Awesome! Let me know if you want this PR merged now or if you'll keep working on it to remove the filter.

parthea:
Thanks @yebrahim ! Yes, please merge it. I'll tackle the other codes tomorrow. Hopefully we can remove the filter next week.

PR #315, opened 2017-03-20
Fix all flake8 warnings. Enable all flake8 codes except E111, E114, E121

parthea:
Enable all flake8 codes on Travis except
- E111 # Indentation is not a multiple of four
- E114 # Indentation is not a multiple of four (comment)
- E121 # Continuation line under-indented for hanging indent

For this PR, I assumed that the number of spaces for indentation will remain at 2.

There are some lines that produced flake8 codes which I could not fix. Those lines have the comment `# noqa` to bypass flake8 testing. We also have the option to ignore certain flake8 codes altogether in case we will have reduced code readability. For example, we may choose to continue using `lambda` in the code rather than `def`. See [Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier.](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)

You can input the prefix of a flake8 code in the `--select` option of the flake8 command. This allowed me to break up the work into separate, more manageable commits based on the flake8 code prefix. For example, the following command will display all E1XX, E2XX and E3XX codes. 

`flake8 --select=E1,E2,E3 --statistics`

Also, while going through flake8 codes I noticed that the following line of code seemed out of place. Should we create a separate ticket for it? 
https://github.com/googledatalab/pydatalab/blob/master/solutionbox/image_classification/mltoolbox/image/classification/_inceptionlib.py#L526

I'm happy to make any changes as required, regardless of how major/minor the change may be. It's a great learning experience for me.


yebrahim:
This is great! Thank you very much for taking the time and effort to do this @parthea.

Yes that line looks like it was left by mistake. I'll let @qimingj or @brandondutra confirm.

yebrahim:
I'll merge this now then we can fix that one line separately.

parthea:
Thanks @yebrahim !

PR #316, opened 2017-03-22
sd analyze interface unit tests for python 2 and 3

brandondutra:
This PR does two things
1) addes unit tests to the SD datalab interface functions just for analyze. The real work of analyze (local_preprocess.py and cloud_preprocess.py) is not tested in this PR
2) makes these tests work in python 2 and 3. It turns out datalab does not support python 3 as is. This is surprising because travis ci run tests in python 3 too, so this shows we have a serious test gap. 

Some notes on file changes:

google/datalab/_job.py: print the stack trace on error, so debugging information is given. 

google/datalab/ml/_dataset.py: python 2 or 3 was complaining that basestring was used before assigned.

google/datalab/ml/_util.py: remove usage of apache_beam, which does not have a python 3 version (see https://cloud.google.com/dataflow/docs/quickstarts/quickstart-python#install-pip-and-the-dataflow-sdk). _util can now be used in python 3 tests.

setup.py: pin scikit-learn to the same version in the docker container.

all other files: get the analyze tests working in python 3


setup.py
```diff
@@ -97,6 +97,7 @@
     'pytz>=2015.4',
     'pyyaml==3.11',
     'requests==2.9.1',
+    'scikit-learn==0.17.1',
     'ipykernel==4.4.1',
     'psutil==4.3.0',
```
parthea: Unrelated to your changes, `'psutil==4.3.0',` appears twice in this list. Could it be fixed in this PR or should we create a separate issue?

solutionbox/structured_data/test_mltoolbox/test_package_functions.py
```diff
@@ -0,0 +1,139 @@
...
+    self.assertIs(core_sd.analyze_async, classdnn.analyze_async)
+    self.assertIs(core_sd.analyze_async, reglin.analyze_async)
+    self.assertIs(core_sd.analyze_async, regdnn.analyze_async)
+
+  def test_analysis_argspec(self):
+    """Test all analyze functions have the same parameters"""
+    if sys.version_info.major > 2:
+      spec_fn = inspect.getfullargspec  # for python 3 compatibility
+    else:
+      spec_fn = inspect.getargspec
```
parthea: Do you think it could be helpful to have a compatibility module pyDatalab? The compatibility module could contain helper function for common python 2/3 compatibility issues. For example:

```
def compat_inspect_getargspec(obj):
	if sys.version_info.major > 2:
		return obj.getfullargspec  # for python 3 compatibility
	else:
		return obj.getargspec(**kw)
```

Then we can use the compatibility functions directly:
```
    self.assertEqual(spec_fn(core_sd.analyze), compat_inspect_getargspec(core_sd.analyze_async))
    self.assertEqual(spec_fn(core_sd.analyze), compat_inspect_getargspec(core_sd._analyze))
```

solutionbox/structured_data/mltoolbox/_structured_data/preprocess/local_preprocess.py
```diff
@@ -135,7 +136,12 @@ def _init_numerical_results():
       json.dumps(numerical_results, indent=2, separators=(',', ': ')))
 
   # Write the vocab files. Each label is on its own line.
-  for name, unique_labels in categorical_results.iteritems():
+  if sys.version_info.major > 2:
+    name_label_itr = categorical_results.items()
+  else:
+    name_label_itr = categorical_results.iteritems()
+
+  for name, unique_labels in name_label_itr:
```
parthea: See my other comment about potentially having a compatibility function so that this could be changed to:

```
  for name, unique_labels in compat_iteritems(categorical_results):
```

solutionbox/structured_data/mltoolbox/_structured_data/_package.py
```diff
@@ -59,15 +59,7 @@ class FileNotFoundError(IOError):
...
-  return Context.default().project_id
-
-
-def _is_in_IPython():
-  try:
-    import IPython # noqa
-    return True
-  except ImportError:
-    return False
+  return Context.default().project_id.decode()
```
parthea: If the `.decode()` is for python compatibility we could build a helper function that will automatically apply the `.decode()` if python 3.

setup.py
brandondutra: done

solutionbox/structured_data/test_mltoolbox/test_package_functions.py
brandondutra: I see the docker container has the python package six, which looks like the official compatibility module. I'll look into using that.

solutionbox/structured_data/mltoolbox/_structured_data/preprocess/local_preprocess.py
brandondutra: used six.iteritems

solutionbox/structured_data/test_mltoolbox/test_package_functions.py
brandondutra: six has a get_function_defaults() function for getting the default values, but I don't see a function for getting all function arguments. 

I pulled out the argspec stuff into a method of the TestFunctionSignature class. It may not be useful to pull  getargspec into a new compatibility datalab module as this test class will likely be the only one that will use it.

solutionbox/structured_data/mltoolbox/_structured_data/_package.py
brandondutra: py2: Context.default().project_id is type str, after decode() type is  unicode.
py3: Context.default().project_id is type bytes, after decode() type is str (which is unicode)

Moved this logic into get_default_project_id() function.

now:
in py2: project_id is type str
in py3: project_id is type str

The problem is that the output of  subprocess.Popen().communicate()  can be anything (say image data), so this isn't really a compatability question of converting str to unicode, but how bytes should be converted to a string type.

google/datalab/ml/_util.py
```diff
@@ -13,10 +13,9 @@
 # limitations under the License.
 
 
-from apache_beam.io import gcsio
-import glob
 from googleapiclient import discovery
-import logging
+from tensorflow.python.lib.io import file_io as tfio
```
qimingj: Can we use tf.gfile, which seems come with public document (https://www.tensorflow.org/api_docs/python/tf/gfile/GFile).

brandondutra: done. Updated it here, but not in the mltoolbox packages.

PR #325, opened 2017-03-27
Update ipykernel dependecy version

yebrahim:
The only clear dependency is that IPython is defined. Exact match is preventing Datalab from updating `ipykernel`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,7 +97,7 @@ for accessing Google's Cloud Platform services such as Google BigQuery.
     'pyyaml==3.11',
     'requests==2.9.1',
     'scikit-learn==0.17.1',
-    'ipykernel==4.4.1',
+    'ipykernel==4.5.2',
     'psutil==4.3.0',
     'jsonschema==2.6.0',
   ],


setup.py
```diff
@@ -97,7 +97,7 @@
     'pyyaml==3.11',
     'requests==2.9.1',
     'scikit-learn==0.17.1',
-    'ipykernel==4.4.1',
+    'ipykernel<=4.5.2',
```
ojarjur: I don't think I want to leave it ambiguous as to which version we install.

Can you explain what goes wrong if this is just `ipykernel==4.5.2`?

yebrahim: Well there's no reason why this package shouldn't work for you if you're installing it on an earlier version of `notebook`, which has an earlier version of `ipykernel`.
However, looking at the dependency of `notebook` on `ipykernel`, I see it's unbound to a certain version, so we should be fine here. Even if you have an older `notebook` version, it's fine that `pydatalab` updates your `ipykernel`. I'll pin it to `4.5.2`.

PR #328, opened 2017-03-29
Use UTC in client so dates and times match UTC as used on the server.

jimmc:
Fixes https://github.com/googledatalab/datalab/issues/1195
Changes client code to show dates and times in UTC, which matches the string conversions that come from the kernel, since the container uses UTC.
See screenshots for details.

date and time tables before:
timedata-before

date and time tables after:
timedata-after

timestamp graph before:
timestamps-graph-before

timestamp graph after:
timestamps-graph-after

date graph before:
date-graph-before

date graph after:
date-graph-after

datetime graph before:
datetimes-graph-before

datetime graph after:
datetimes-graph-after

time graph before (no change, this one was correct):
times-graph-before

time graph after
times-graph-after

### DIFF
--- a/google/datalab/notebook/static/charting.ts
+++ b/google/datalab/notebook/static/charting.ts
@@ -507,6 +507,7 @@ module Charting {
             const v = new Date(rows[row].c[col].v);
             if (v.getTime() % 1000 != 0) {
               formatter = timestampWithFractionalSecondsFormatter;
+              break;
             }
           }
           for (let row = 0; row < rows.length; row++) {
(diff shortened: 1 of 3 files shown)


google/datalab/notebook/static/charting.ts
```diff
@@ -438,22 +438,82 @@ module Charting {
...
               });
           }
+        } else if (data.cols[col].type == 'timestamp') {
+          data.cols[col].type = 'datetime';
+          // Run through all the dates to determine how to format them.
+          let formatter = timestampFormatter;
+          for (let row = 0; row < rows.length; row++) {
+            const v = new Date(rows[row].c[col].v);
+            if (v.getTime() % 1000 != 0) {
+              formatter = timestampWithFractionalSecondsFormatter;
```
qimingj: You can break out of the for loop now after you determined timestampWithFractionalSecondsFormatter is needed?

jimmc: Yup. Done.

Issue #330, opened 2017-03-30
Running tests with no gcloud project set fails


craigcitro:
I was running this locally, and hit the following:

```
======================================================================
FAIL: test_unsupported_schema (test_mltoolbox.test_package_functions.TestAnalyze)
Test supported schema values.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/google/home/craigcitro/gh/pydatalab/solutionbox/structured_data/test_mltoolbox/test_package_functions.py", line 113, in test_unsupported_schema
    job.fatal_error.message)
AssertionError: 'Schema contains an unsupported type bytes.' not found in 'No project ID found. Perhaps you should set one by running"%datalab project set -p <project-id>" in a code cell.'

----------------------------------------------------------------------
```

Setting a project ID (via `gcloud config set project myproject`) fixed the issue.

/cc @brandondutra

brandondutra:
notes to self: when testing the interface functions, pass in a dummy project id and use it over the default, which calls gcloud.

PR #331, opened 2017-03-30
Set up tox, and use it for local testing and Travis.

craigcitro:
[tox](https://tox.readthedocs.io/) is a Python test automation tool: it makes it easy to automate running the tests for your project in several environments. Under the hood, it just uses virtualenv to manage installed deps.

If you have python2.7 and python3.5 installed, simply running `tox` at the root of the repo will now run tests in both versions, as well as `flake8`.

This PR makes two changes:

* first, convert the existing `.travis.yml` into `tox.ini`, setting up three tox environments based on the configuration that was already there. (Relatedly, I added the `.tox` directory to `.gitignore`).

* Now that #330 is fixed (wooo @brandondutra), I switched the `.travis.yml` over to use `tox`. Now there are separate travis environments for each tox environment, which has two advantages: (1) these three all run in parallel, speeding up builds, and (2) build logs are separated, making it faster to figure out what caused a given failure.

PTAL @yebrahim @brandondutra

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,32 +1,22 @@
-sudo: required
-
-dist: trusty
-
 language: python
+dist: trusty
+sudo: false
 
-python:
-  - 2.7
-  - 3.5
+matrix:
+  include:
+  - python: 2.7
+    env: TOX_ENV=py27
+  - python: 3.5
+    env: TOX_ENV=py35
+  - python: 2.7
+    env: TOX_ENV=flake8
 
 before_install:
-  - sudo apt-get install -y python-setuptools
   - npm install -g typescript
   - tsc --module amd --noImplicitAny --outdir datalab/notebook/static datalab/notebook/static/*.ts
-  - pip install -U pip
-  - pip install -U setuptools
-  - if [[ $TRAVIS_PYTHON_VERSION == 2* ]]; then pip install -U google-cloud-dataflow==0.5.5; fi
-  - pip install .
-  - pip install solutionbox/structured_data/
-  - pip install solutionbox/image_classification/  
-  - pip install -U tensorflow==1.0.0
-  - pip install -U protobuf==3.1.0  #v3.2.0 has the "python: double free or corruption" bug.
-  - pip install flake8
-
-before_script:
-  - flake8 .
-
+  - pip install --upgrade pip setuptools tox coveralls
+  
 script:
-  - python ./tests/main.py
-  - python ./legacy_tests/main.py
+  - tox -e $TOX_ENV
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,11 @@ matrix:
 before_install:
   - npm install -g typescript
   - tsc --module amd --noImplicitAny --outdir datalab/notebook/static datalab/notebook/static/*.ts
-  - pip install --upgrade pip setuptools tox coveralls
+  # We use tox for actually running tests.
+  - pip install --upgrade pip tox
   
 script:
+  # tox reads its configuration from tox.ini.
   - tox -e $TOX_ENV
 
 
--- a/tox.ini
+++ b/tox.ini
@@ -11,8 +11,10 @@ skip_missing_interpreters = true
 #
 # tox always installs the current package, so there's no need to list it here.
 deps = tensorflow==1.0.0
-       #Note: protobuf 3.2.0 has the "python: double free or corruption" bug.
-       protobuf==3.1.0 
+       # Note: protobuf 3.2.0 has the "python: double free or corruption" bug.
+       protobuf==3.1.0
+       # Dropping this seems to cause problems with conda in some cases.
+       scipy
        solutionbox/structured_data/
        solutionbox/image_classification/
 commands =
--- a/tox.ini
+++ b/tox.ini
@@ -15,9 +15,6 @@ commands =
 deps = {[testenv]basedeps}
        google-cloud-dataflow==0.5.5
 
-[flake8]
-exclude = .tox,.git,./*.egg,build,.cache,env,__pycache__,docs
-
 [testenv:flake8]
-commands = flake8
+commands = flake8 --exclude=.tox,.git,./*.egg,build,.cache,env,__pycache__,docs
 deps = flake8
--- a/tox.ini
+++ b/tox.ini
@@ -1,18 +1,27 @@
 [tox]
+# By default, we want to run tests for Python 2.7, Python 3.5, and run our
+# flake8 checks.
 envlist = py27,py35,flake8
+# If an interpreter is missing locally, skip it.
+skip_missing_interpreters = true
 
 [testenv]
-basedeps = tensorflow==1.0.0
-           protobuf==3.1.0
-           solutionbox/structured_data/
-           solutionbox/image_classification/
-deps = {[testenv]basedeps}
+# pydatalab doesn't require users to have these dependencies installed, but we
+# need them to run our tests suite.
+#
+# tox always installs the current package, so there's no need to list it here.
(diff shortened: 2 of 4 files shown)


brandondutra:
sorry, what is tox and why do we need it?

craigcitro:
@brandondutra Ah -- yes, I had chatted with @yebrahim about `tox`, but realized that I totally failed to include that info in the PR. 😊 

Updated the commit description, updating PR description to match in a sec.

If there's a good place to add more docs about this, happy to do so.

.travis.yml
```diff
@@ -1,32 +1,22 @@
...
+  include:
+  - python: 2.7
+    env: TOX_ENV=py27
+  - python: 3.5
+    env: TOX_ENV=py35
+  - python: 2.7
+    env: TOX_ENV=flake8
 
 before_install:
-  - sudo apt-get install -y python-setuptools
```
yebrahim: it's great that we're cleaning up these obsolete requirements too!

tox.ini
```diff
@@ -0,0 +1,23 @@
+[tox]
```
yebrahim: it'd be great if you can add a readme file in the tests directory to list the local testing steps (I'm assuming just install and run tox?). that's one of the things missed by newcomers to the project. @jimmc

tox.ini
```diff
@@ -0,0 +1,23 @@
...
+deps = {[testenv]basedeps}
+commands =
+  python ./tests/main.py
+  python ./legacy_tests/main.py
+
+[testenv:py27]
+deps = {[testenv]basedeps}
+       google-cloud-dataflow==0.5.5
+
+[flake8]
```
yebrahim: should this be `testenv:flake8` for consistency?

craigcitro: So these were actually the args *to* `flake8` when invoked, but that was a little mysterious.

For now, I just went ahead and inlined those args -- we can get fancier later as needed.

tox.ini
craigcitro: Good call, though I did it with a twist -- I went ahead and modified `CONTRIBUTING.md`, since (presumably) anyone contributing to this package is going to want to know how to write tests.

tox.ini
yebrahim: I see. that wasn't easy to infer, yes this is much clearer.

.travis.yml
```diff
@@ -1,32 +1,22 @@
...
-
-before_script:
-  - flake8 .
-
+  - pip install --upgrade pip setuptools tox coveralls
+  
 script:
-  - python ./tests/main.py
-  - python ./legacy_tests/main.py
+  - tox -e $TOX_ENV
```
brandondutra: add comment on how it is finding the tests to run.

.travis.yml
```diff
@@ -1,32 +1,22 @@
...
-  - pip install solutionbox/structured_data/
-  - pip install solutionbox/image_classification/  
-  - pip install -U tensorflow==1.0.0
-  - pip install -U protobuf==3.1.0  #v3.2.0 has the "python: double free or corruption" bug.
-  - pip install flake8
-
-before_script:
-  - flake8 .
-
+  - pip install --upgrade pip setuptools tox coveralls
```
brandondutra: why do we have two sets of dependencies 
1) install things in this list
2) install things in tox.ini

Can we have just one list?

tox.ini
```diff
@@ -0,0 +1,20 @@
+[tox]
+envlist = py27,py35,flake8
+
```
brandondutra: can we add some comments on what is going on for non-tox experts?

tox.ini
```diff
@@ -0,0 +1,20 @@
...
+[testenv]
+basedeps = tensorflow==1.0.0
+           protobuf==3.1.0
+           solutionbox/structured_data/
+           solutionbox/image_classification/
```
brandondutra: so datalab is no longer being installed (pip install .). Maybe add a comment?

.travis.yml
craigcitro: Done.

brandondutra:
Travis is going to run the tests in a container. The container  sets python in a virtualenv, and then it runs tox which automatically makes another virtualenv? 

By adding another layer of complexity to our tests we gain:
1) flake8 is now its own test
2) local dev testing of python 2/3 without having two windows open?

The Travis build already ran in parallel and build logs were separated. Right? So I don't understand the advantages you list in the PR description.

tox.ini
craigcitro: `tox` always installs the current package -- added a note to that effect.

.travis.yml
craigcitro: This list is now just `tox`, which is as minimal as it can get.

tox.ini
```diff
@@ -0,0 +1,28 @@
...
+# If an interpreter is missing locally, skip it.
+skip_missing_interpreters = true
+
+[testenv]
+# pydatalab doesn't require users to have these dependencies installed, but we
+# need them to run our tests suite.
+#
+# tox always installs the current package, so there's no need to list it here.
+deps = tensorflow==1.0.0
+       protobuf==3.1.0
```
brandondutra: you lost the comment on why 3.2.0 is bad.

craigcitro:
The main goal is really to automate as much as possible: I don't think there's any advantage to manually managing virtualenvs, and this makes it easier for newcomers to figure out how to run tests for the project (eg me, earlier today). `tox` seems to be the common answer for most popular Python projects nowadays.

tox also has the advantage that it'll work for other testing and release steps, eg building documentation: you add a new entry in `tox.ini`, and then `tox -e docs` builds for developers, it can run on travis automatically for incoming PRs, and it's just one more step to have it push those docs to the right branch on release.

tox.ini
craigcitro: Done -- LMK if it's still unclear.

tox.ini
craigcitro: Yes! Good catch.

craigcitro:
@brandondutra Thinking about this a bit more, maybe I understand your concern -- is the issue that we're moving our deps from one opaque place (the Travis-CI config) to another opaque place (the tox config)?

If so, I'd also be happy to split this differently: move the deps to `requirements.txt` and `test-requirements.txt`, and have tox reference those. Then users like me can just type `tox`; for folks like you, who maybe don't feel the need to add another layer into their toolchain, you're still ahead of where we started: `pip install -f test-requirements.txt` gets you everything you need to run locally, and you're off to the races.

Does that sound better?

brandondutra:
The requirements thing is fine, no need to make changes. 

I installed tox today and played with the new build setup. I could not get the tox python 3 test working in my conda python 2 env, which I guess makes sense. But I was able to get both tox python tests working in my conda python 3 env, which is cool. 

In my python 2 and 3 env, the tox python 2 test fails saying I don't have scipy installed. But I have scipy in both conda envs. When tox runs the python 2 tests, it is not listed as installed, but is listed in the tox python 3 test. Adding scipy to setup.py does not fix my scipy error, but adding scipy to tox.ini did fix it. 

So maybe add scipy as a dep in tox?

craigcitro:
Added scipy -- thanks @brandondutra!

Do I need to squash, or will you guys do that on merge?

brandondutra:
You can just click the green squash and merge button!

craigcitro:
screenshot 2017-03-31 at 10 57 31 am

😢

qimingj:
Omar controls permissions.

yebrahim:
I'm guessing the reason we need the `scipy` dep is that the mltoolbox packages are installed before datalab? datalab already lists `seaborn`, which takes `scipy` as a dependency. I think this is fine for now, but we should be aware of the install order.

Issue #332, opened 2017-03-31
structure data analyze function does not use project id


brandondutra:
it does not use project id that is passed in, and instead checks the default project id.

PR #333, opened 2017-03-31
make the tests work without a project

### LINKED ISSUE #330: Running tests with no gcloud project set fails

brandondutra:
...without a project set in gcloud

analyze does not use the project_id that is passed in! This is documented in #332 This will change if we move to tensorflow_transform. 

Fixes #330

### DIFF
--- a/solutionbox/structured_data/mltoolbox/_structured_data/_package.py
+++ b/solutionbox/structured_data/mltoolbox/_structured_data/_package.py
@@ -189,14 +189,17 @@ def _analyze(output_dir, dataset, cloud=False, project_id=None):
                                            prefix='schema')
     file_io.write_string_to_file(schema_file_path, json.dumps(dataset.schema))
 
+    #TODO(brandondutra) use project_id in the local preprocess function.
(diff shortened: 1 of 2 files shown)


craigcitro:
Thanks for the quick turnaround @brandondutra!

PR #334, opened 2017-03-31
Fix a bug in `%bq query` expansion with UDFs.

craigcitro:
Currently, attempting to call `%bq query --udfs <some udf object>` will fail
with a stack trace like

```
TypeErrorTraceback (most recent call last)
<ipython-input-27-0f9364e227db> in <module>()
----> 1 extract_params_query.sql()

/tmp/tmpZ1tEW3/dist-packages/google/datalab/bigquery/_query.pyc in sql(self)
    183   def sql(self):
    184     """ Get the SQL for the query. """
--> 185     return self._expanded_sql()
    186
    187   @property

/tmp/tmpZ1tEW3/dist-packages/google/datalab/bigquery/_query.pyc in _expanded_sql(self, sampling)
    148
    149     if udfs:
--> 150       expanded_sql += '\n'.join([udfs[udf]._expanded_sql() for udf in udfs])
    151       expanded_sql += '\n'
    152

TypeError: list indices must be integers, not tuple
```

The issue here is a mismatch between use of `query._udfs`, which is a list of
`(name, udf)` tuples, and `query.udfs`, which is `dict(query._udfs)`.

This occurs in the [tutorials](https://github.com/googledatalab/notebooks/blob/master/tutorials/BigQuery/UDFs%20in%20BigQuery.ipynb).

Switching to tuples everywhere fixes this, as verified by the test.

### DIFF
--- a/google/datalab/bigquery/_query.py
+++ b/google/datalab/bigquery/_query.py
@@ -65,7 +65,7 @@ class Query(object):
           if value is None:
             raise Exception('Cannot find object %s' % item)
 
-        # for a dictionary of objects, each pair must be a string an object of the expected type
+        # for a dictionary of objects, each pair must be a string and object of the expected type
         elif isinstance(obj_container, dict):
           value = obj_container[item]
           if not isinstance(value, obj_type):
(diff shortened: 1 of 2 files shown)


craigcitro:
PTAL @yebrahim

craigcitro:
@yebrahim to be fair, I don't get the credit for spotting it -- that's all the automation. 😁

PR #335, opened 2017-03-31
Added travis build badge to README

yebrahim:
Added travis build badge.


yebrahim:
Good idea, added.

README.md
```diff
@@ -1,4 +1,4 @@
-# datalab
+# datalab [![Build Status](https://travis-ci.org/googledatalab/pydatalab.svg?branch=master)](https://travis-ci.org/googledatalab/pydatalab) [![PyPI Package](https://img.shields.io/pypi/v/datalab.svg)](https://pypi.python.org/pypi/datalab/1.0.1)
```
parthea: I noticed that version 1.0.1 is hardcoded in the pypi link. Could we link directly to https://pypi.python.org/pypi/datalab/ instead?

yebrahim: Isn't this a static image though? I mean it wouldn't update itself when a new pypi version is pushed right?

parthea: >Isn't this a static image though? I mean it wouldn't update itself when a new pypi version is pushed right?

Correct. I believe that the image at https://img.shields.io/pypi/v/datalab.svg will update automatically. 

I have a concern about the link. If I understand correctly, the link appears to be hardcoded to 1.0.1 which doesn't seem correct.

yebrahim: I see, I didn't know it would update automatically. Fixed.

Issue #329, opened 2017-03-29
Diagnose, benchmark and provide guidance for loading large dataframe from BigQuery


Di-Ku:
Customer query (via Tahir F.)

Do we have any kind of benchmarks / recommendations for the GCE set-up for the amount of data that would be brought into a pandas dataframe? 
His question is as follows:
From my perspective, could you advise me the appropriate spec of GCE?
We grade up the GCE spec and it seems to use only 2% of CPU but it takes 5mins to handle 500,000rows data in pandas.
 
Do you have any idea to improve the performance of datalab?
Does it relate to network or disk issue?

parthea:
I've also noticed that loading a large DataFrame from BigQuery is slow. My machine is outside of GCE so I understand that there will be significant delays in downloading data.  Still, my feeling (but still need to confirm) is that the following process is quicker compared to bq.Query(<sqlModule>).to_dataframe(dialect='standard')` despite #10 . 
- Run  `bq.Query(<sqlModule>).extract(<gcs object path>)` 
- Then download the file using `storage.Bucket(<my bucket>).item(<item>).read_from()`
- Followed by `pd.read_csv(<path to csv file>)`


Some initial profiling data:

Using `bq.Query().extract()`: Total Time => ~2 minutes , 39 seconds
```
bq.Query(<sqlModule>).extract(<gcs path>), dialect='standard', use_cache=False)
```
```
         167823 function calls (167674 primitive calls) in 158.961 seconds

   Ordered by: internal time
   List reduced from 431 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       17   85.070    5.004   85.070    5.004 {time.sleep}
    15523   65.955    0.004   65.955    0.004 {method 'read' of '_ssl._SSLSocket' objects}
       23    5.371    0.234    5.371    0.234 {_socket.getaddrinfo}
       23    1.411    0.061    1.411    0.061 {method 'do_handshake' of '_ssl._SSLSocket' objects}
       23    0.711    0.031    0.711    0.031 {method 'connect' of '_socket.socket' objects}
        2    0.263    0.132    0.263    0.132 {method 'read' of 'file' objects}
       23    0.064    0.003    0.064    0.003 {method 'load_verify_locations' of '_ssl._SSLContext' objects}
      704    0.018    0.000   66.000    0.094 socket.py:406(readline)
    15523    0.009    0.000   65.976    0.004 ssl.py:707(recv)
[...]

PR #339, opened 2017-04-05
Table.to_dataframe optimization fixes

yebrahim:
- Use large page size for downloading entire tables
- Concatenate paged dataframes instead of appending

Follow up from the discussion on https://github.com/googledatalab/pydatalab/issues/329.
Replaces https://github.com/googledatalab/pydatalab/pull/220

### DIFF
--- a/google/datalab/bigquery/_table.py
+++ b/google/datalab/bigquery/_table.py
@@ -103,8 +103,11 @@ class Table(object):
   # Allowed characters in a BigQuery table column name
   _VALID_COLUMN_NAME_CHARACTERS = '_abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789'
 
-  # When fetching table contents, the max number of rows to fetch per HTTP request
+  # When fetching table contents for a range or iteration, use a small page size per request
   _DEFAULT_PAGE_SIZE = 1024
+  # When fetching the entire table, use the maximum number of rows. The BigQuery service
+  # is likely to return less rows than this if their encoded JSON size is larger than 10MB
+  _MAX_PAGE_SIZE = 100000
 
   # Milliseconds per week
   _MSEC_PER_WEEK = 7 * 24 * 3600 * 1000
@@ -595,20 +598,24 @@ class Table(object):
     Returns:
       A Pandas dataframe containing the table data.
     """
-    fetcher = self._get_row_fetcher(start_row=start_row, max_rows=max_rows)
+    fetcher = self._get_row_fetcher(start_row=start_row,
+                                    max_rows=max_rows,
+                                    page_size=self._MAX_PAGE_SIZE)
     count = 0
     page_token = None
-    df = None
+
+    # Collect results of page fetcher in separate dataframe objects, then
+    # concatenate them to reduce the amount of copying
+    df_list = []
+
     while True:
       page_rows, page_token = fetcher(page_token, count)
       if len(page_rows):
         count += len(page_rows)
-        if df is None:
-          df = pandas.DataFrame.from_records(page_rows)
-        else:
-          df = df.append(page_rows, ignore_index=True)
+        df_list.append(pandas.DataFrame.from_records(page_rows))
       if not page_token:
         break
+    df = pandas.concat(df_list, ignore_index=True, copy=False)
 
     # Need to reorder the dataframe to preserve column ordering
     ordered_fields = [field.name for field in self.schema]
--- a/google/datalab/bigquery/_table.py
+++ b/google/datalab/bigquery/_table.py
@@ -106,7 +106,7 @@ class Table(object):
   # When fetching table contents for a range or iteration, use a small page size per request
   _DEFAULT_PAGE_SIZE = 1024
   # When fetching the entire table, use the maximum number of rows. The BigQuery service
-  # is likely to return less rows than this if their encoded JSON size is larger than 10MB
+  # will always return fewer rows than this if their encoded JSON size is larger than 10MB
   _MAX_PAGE_SIZE = 100000
 
   # Milliseconds per week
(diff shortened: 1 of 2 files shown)


google/datalab/bigquery/_table.py
```diff
@@ -103,8 +103,11 @@ class Table(object):
   # Allowed characters in a BigQuery table column name
   _VALID_COLUMN_NAME_CHARACTERS = '_abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789'
 
-  # When fetching table contents, the max number of rows to fetch per HTTP request
+  # When fetching table contents for a range or iteration, use a small page size per request
   _DEFAULT_PAGE_SIZE = 1024
+  # When fetching the entire table, use the maximum number of rows. The BigQuery service
+  # is likely to return less rows than this if their encoded JSON size is larger than 10MB
```
craigcitro: grammar nit: `less rows` -> `fewer rows`

Also, I think it's not just likely -- BigQuery *will* return fewer rows than this.

google/datalab/bigquery/_table.py
```diff
@@ -595,20 +598,24 @@ def to_dataframe(self, start_row=0, max_rows=None):
...
+
     while True:
       page_rows, page_token = fetcher(page_token, count)
       if len(page_rows):
         count += len(page_rows)
-        if df is None:
-          df = pandas.DataFrame.from_records(page_rows)
-        else:
-          df = df.append(page_rows, ignore_index=True)
+        df_list.append(pandas.DataFrame.from_records(page_rows))
```
craigcitro: As one more potential speed comparison: did we consider just collating the list of rows (not slices of the final `DataFrame`), and then only creating a single `DataFrame` at the end? That is, something like

```python
rows = []
while True:
  ...
  if len(page_rows):
    rows.extend(page_rows)
  ...
df = pd.DataFrame.from_records(rows)
```

google/datalab/bigquery/_table.py
```diff
@@ -103,8 +103,11 @@ class Table(object):
   # Allowed characters in a BigQuery table column name
   _VALID_COLUMN_NAME_CHARACTERS = '_abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789'
 
-  # When fetching table contents, the max number of rows to fetch per HTTP request
+  # When fetching table contents for a range or iteration, use a small page size per request
   _DEFAULT_PAGE_SIZE = 1024
+  # When fetching the entire table, use the maximum number of rows. The BigQuery service
+  # is likely to return less rows than this if their encoded JSON size is larger than 10MB
+  _MAX_PAGE_SIZE = 100000
```
craigcitro: Any reason to make this a constant instead of just an optional arg to `to_dataframe`?

google/datalab/bigquery/_table.py
```diff
@@ -103,8 +103,11 @@ class Table(object):
   # Allowed characters in a BigQuery table column name
   _VALID_COLUMN_NAME_CHARACTERS = '_abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789'
 
-  # When fetching table contents, the max number of rows to fetch per HTTP request
+  # When fetching table contents for a range or iteration, use a small page size per request
```
craigcitro: Not actually related to this CL: I have to admit, I feel a little confused by `1024` here.

If the goal really is low latency for operations like "let me see a sample of my dataframe", then 100 rows is already more than enough, and should be faster yet than 1024.

If the goal is to minimize traffic, I feel like we should just make this 100k -- we already know that BQ is going to cap us at ~10MB of data.

You could totally shut me up on questions like this with some nice tables about speeds for smallish numbers of rows. 😁

google/datalab/bigquery/_table.py
yebrahim: Not necessarily, if the row is very small including the header, you can fit more than 100,000 rows in 10MB.

google/datalab/bigquery/_table.py
yebrahim: It is an optional arg to `to_dataframe`, this is its default value, which is good to keep it here as a clear static constant than a hardcoded number.

yebrahim:
I agree about tests, will work on them in this PR.

google/datalab/bigquery/_table.py
craigcitro: OK, I'm nitpicking, but the sentence here is "BQ is likely to return fewer rows than this *if their encoded JSON size is less than 10MB*." If that second part is true, then it's not just likely, it's guaranteed, right? :wink:

yebrahim: Oh I see, you're right. Fixed. :)

google/datalab/bigquery/_table.py
yebrahim: I experimented with this a little, but I'm not seeing any speedup. It looks like creating a dataframe out of a list is a cheap operation, and extending large lists might sometimes result in copying so that might be offsetting any benefit from using one big list.

google/datalab/bigquery/_table.py
yebrahim: Sure, here are some numbers: :)

| Page size | Fetch time |
| ------------- | ------------- |
| 50 | ~160ms |
| 100 | ~180ms |
| 1024 | ~700ms |
| 10,000 | ~5s |

So we need to minimize latency, because going the 100k route will have the users waiting for ~5 seconds to get their first page (even though they probably won't wait for another fetch). This also applies if you want to use the table iterator to get rows 1000 to 1020 for example.

craigcitro: So based on that data, should this constant be `100` instead of `1024`?

(Again, probably worth splitting off into a different issue, since it's largely orthogonal.)

google/datalab/bigquery/_table.py
craigcitro: sgtm.

craigcitro:
I'm happy with all the existing bits modulo tests. :grinning:

yebrahim:
I added a simple unit test to validate the page size. I don't think we should have a test that times calls to the service as part of the unit tests, this seems like it should be part of a benchmark suite that measures several aspects of the APIs. Thoughts are welcome.

craigcitro:
I agree on the separate benchmark -- maybe worth filing an issue?

yebrahim:
Sure, opened https://github.com/googledatalab/pydatalab/issues/342.

Issue #338, opened 2017-04-04
Table.extract_async in avro format not working


JokullJohanns:
Trying to run the following code

```
project = Context.default().project_id
bucket_name = 'jokull'
bucket_path = 'gs://{}/{}'.format(bucket_name, "datasets/dataset_{}".format(date))

train_table = bq.Table("*myproject*.*mydataset*.Trainingset_{}".format(date))
train_bucket_object = bucket_path + '/Trainingset.csv'
train_table.extract_async(destination=train_bucket_object,format=u"avro")
```

but I'm getting the following error...

```
AttributeErrorTraceback (most recent call last)
<ipython-input-39-12fc3977cc46> in <module>()
      5 train_table = bq.Table("*myproject*.*mydataset*.Trainingset_{}".format(date))
      6 train_bucket_object = bucket_path + '/Trainingset.csv'
----> 7 train_table.extract_async(destination=train_bucket_object,format=u"avro")

/usr/local/lib/python2.7/dist-packages/google/datalab/bigquery/_table.pyc in extract_async(self, destination, format, csv_delimiter, csv_header, compress)
    409       return self._init_job_from_response(response)
    410     except Exception as e:
--> 411       raise google.datalab.utils.JobError(location=traceback.format_exc(), message=str(e),
    412                                           reason=str(type(e)))
    413 

AttributeError: 'module' object has no attribute 'JobError'
```

Replaced *myproject*.*mydataset* with the names of my actual tables.

When I look at the source code I don't see that avro format is supported although the documentation states that it does.

Would be really nice if you could have a look at this and let me know.

Best regards,
Jökull

JokullJohanns:
I found a solution that is working for now...

```
def store_avro_to_gcs(T, bucket_object): #T=bq.Table()
    _ENDPOINT = 'https://www.googleapis.com/bigquery/v2'
    _JOBS_PATH = '/projects/%s/jobs/%s'

    url = _ENDPOINT + (_JOBS_PATH % (T._name_parts.project_id, ''))
    data = {
            'kind': 'bigquery#job',
            'configuration': {
                'extract': {
                    'sourceTable': {
                        'projectId': T._name_parts.project_id,
                        'datasetId': T._name_parts.dataset_id,
                        'tableId': T._name_parts.table_id,
                    },
                    'destinationUris': [bucket_object],
                    'destinationFormat': u'AVRO',
                }
            }
        }
[...]

PR #345, opened 2017-04-06
Fix bad default for csv delimiter

yebrahim:
Fixes https://github.com/googledatalab/pydatalab/issues/338.

For AVRO data types, no csv delimiter should be passed.
Also fixes a bad module path for `JobError`.

### DIFF
--- a/google/datalab/bigquery/_table.py
+++ b/google/datalab/bigquery/_table.py
@@ -414,7 +414,7 @@ class Table(object):
       return self._init_job_from_response(response)
     except Exception as e:
       raise google.datalab.JobError(location=traceback.format_exc(), message=str(e),
-                                          reason=str(type(e)))
+                                    reason=str(type(e)))
 
   def extract(self, destination, format='csv', csv_delimiter=None, csv_header=True, compress=False):
     """Exports the table to GCS; blocks until complete.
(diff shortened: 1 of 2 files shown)


google/datalab/bigquery/_table.py
```diff
@@ -388,7 +388,7 @@ def _init_job_from_response(self, response):
       job = _job.Job(job_id=response['jobReference']['jobId'], context=self._context)
     return job
 
-  def extract_async(self, destination, format='csv', csv_delimiter=',', csv_header=True,
+  def extract_async(self, destination, format='csv', csv_delimiter=None, csv_header=True,
```
qimingj: I am thinking would it be better if we set it to None if format != 'csv'? Setting default to None seems not clear (docstring says defaults to ',', but seems not true from the function interface).

yebrahim: The docstring is still correct, the default is `","` only in the case of csv exports.

I don't think we should do opaque changes, the user should be responsible for specifying a bad parameter.

qimingj: "defaults to ','" is different from "if None and format is 'csv', ',' is used". Perhaps I am nitpicking here. :)

It seems BigQuery should ignore csv_delimiter if format is not 'csv'. We could always pass None to BigQuery if format is not 'csv' to fix that. I am okay with either.

PR #347, opened 2017-04-09
Fix a few formatting issues in README.

craigcitro:
I was looking at the readme and noticed that there are a few newlines breaking
links; I also fixed a few small spelling/grammar issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,22 @@
 # datalab [![Build Status](https://travis-ci.org/googledatalab/pydatalab.svg?branch=master)](https://travis-ci.org/googledatalab/pydatalab) [![PyPI Package](https://img.shields.io/pypi/v/datalab.svg)](https://pypi.python.org/pypi/datalab)
 
-Google datalab Python package. Used in [Google Cloud Datalab]
-(https://github.com/GoogleCloudPlatform/datalab) and can be used
-in [Jupyter Notebook](http://jupyter.org/).
+[Google Cloud Datalab](https://cloud.google.com/datalab/) Python package. Used in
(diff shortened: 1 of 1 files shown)


craigcitro:
PTAL @yebrahim

README.md
```diff
@@ -1,21 +1,22 @@
...
+See
+[https://github.com/googledatalab/notebooks](https://github.com/googledatalab/notebooks)
 for samples of using this package.
 
 ## Prerequisites
 
-You will need the [Typscript compiler](https://www.typescriptlang.org/) installed.
-In future we should be installable from PyPI.
+You will need the [Typescript compiler](https://www.typescriptlang.org/)
+installed. In the future we should be installable from PyPI.
```
yebrahim: You can take out the part about PyPI, we're already there.

craigcitro: Done.

craigcitro:
Updated -- PTAL.

I also realized two things, which probably warrant a separate PR:

1. I suspect we want to update from `datalab.notebook` to `google.datalab.notebook` in the install scripts.

1. It would be easy to unify those two scripts, since they only differ in one flag, and that could be easily detected as `$VIRTUAL_ENV`.

One last question, again not for this PR: why `datalab` instead of `google-datalab` as the PyPI package name? (Most Google packages tend to have the prefix.)

yebrahim:
I think we should just remove those two scripts, and recommend users do a pip install followed by the nbextension install command, optionally with the `--sys-prefix` argument, all in the README.

I'm not sure how the name was chosen, but it's tough to change that now. :)

Issue #349, opened 2017-04-12
google.datalab.bigquery.Table(...).schema not JSON serializable


jankrynauw:
Looking at the file `schema` property in row 651 of this file: https://github.com/googledatalab/pydatalab/blob/master/google/datalab/bigquery/_table.py

Here is an example of using the `schema` property to retrieve the schema of a table:

<img width="285" alt="screen shot 2017-04-12 at 10 00 38 pm" src="https://cloud.githubusercontent.com/assets/18016598/24976901/84b3d370-1fcb-11e7-9c21-0705763febd2.png">

Using this schema variable in another function which requires a schema as input parameter gives an error: `..is not JSON serializable`:

<img width="739" alt="screen shot 2017-04-12 at 10 29 56 pm" src="https://cloud.githubusercontent.com/assets/18016598/24978073/a09400e8-1fcf-11e7-92bf-ad4740af5356.png">

When I update the schema manually to the following it works: 

<img width="317" alt="screen shot 2017-04-12 at 10 07 54 pm" src="https://cloud.githubusercontent.com/assets/18016598/24977196/92927e78-1fcc-11e7-9acf-adb2cc0a05b3.png">

Is the intention behind the Schema property not there to automatically create a schema from an existing BigQuery table which one could use to populate some of the functions of the ml.toolbox?

chmeyers:
This is because Table.schema() returns a Schema object, but ml.CsvDataSet requires a json schema.  We should either have a way to convert them, or make it so that ml.CsvDataSet accepts a Schema object.

yebrahim:
Yes this is a bug. `CsvDataSet` should be able to accept a `Schema` object as well as a json schema. I'll send a PR to fix this soon.

For now to unblock yourself, you can use the private property `_bq_schema` on the `Schema` object to get the json schema. This should not be your long term solution though, as private properties can change, so please keep that in mind.

PR #350, opened 2017-04-12
Fix CsvDataSet not taking bq.Schema object

yebrahim:
Fixes https://github.com/googledatalab/pydatalab/issues/349.

### DIFF
--- a/google/datalab/ml/_dataset.py
+++ b/google/datalab/ml/_dataset.py
@@ -37,7 +37,7 @@ class CsvDataSet(object):
     Args:
       file_pattern: A list of CSV files. or a string. Can contain wildcards in
         file names. Can be local or GCS path.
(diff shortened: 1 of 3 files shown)


google/datalab/ml/_dataset.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, file_pattern, schema=None, schema_file=None):
     Args:
       file_pattern: A list of CSV files. or a string. Can contain wildcards in
         file names. Can be local or GCS path.
-      schema: A BigQuery schema object in the form of
+      schema: A BigQuery Schema object or a json schema in the form of
```
qimingj: "A google.datalab.bigquery.Schema object or..." so make it super clear?

yebrahim: Sure.

PR #351, opened 2017-04-12
sd package: all categorical columns are multi-label columns

brandondutra:
This means that all string columns are split on space. Also, there
is no difference between categorical columns, multi-label categorical
columns, and text columns.

local/cloud preprocessing updated to build vocabs after splitting. We
 now output the count of each label in a column. This might allow
for tfidf to be supported. It is useful on its own as an analysis check.

### DIFF
--- a/solutionbox/structured_data/mltoolbox/_structured_data/preprocess/local_preprocess.py
+++ b/solutionbox/structured_data/mltoolbox/_structured_data/preprocess/local_preprocess.py
@@ -20,6 +20,7 @@ from __future__ import unicode_literals
 
 import argparse
 import collections
+import csv
 import json
 import os
 import six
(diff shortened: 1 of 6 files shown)


solutionbox/structured_data/mltoolbox/_structured_data/preprocess/local_preprocess.py
```diff
@@ -92,20 +93,25 @@ def _init_numerical_results():
...
-        parsed_line = dict(zip(header, line.strip().split(',')))
+      for line in csv.reader(f):
+        parsed_line = dict(zip(header, line))
 
         for col_schema in schema_list:
           col_name = col_schema['name']
           col_type = col_schema['type']
           if col_type.lower() == 'string':
-            categorical_results[col_name].update([parsed_line[col_name]])
+            split_strings = parsed_line[col_name].split(' ')
```
qimingj: Do you need to filter out empty strings, in case the original string is " good morning "?

solutionbox/structured_data/mltoolbox/_structured_data/preprocess/local_preprocess.py
```diff
@@ -137,8 +143,16 @@ def _init_numerical_results():
       json.dumps(numerical_results, indent=2, separators=(',', ': ')))
 
   # Write the vocab files. Each label is on its own line.
-  for name, unique_labels in six.iteritems(categorical_results):
-    labels = '\n'.join(list(unique_labels))
+  for name, label_count in six.iteritems(categorical_results):
+    # Lables is now the srting:
```
qimingj: typo.

brandondutra: done

solutionbox/structured_data/mltoolbox/_structured_data/preprocess/local_preprocess.py
brandondutra: yea, I thought about this too, I don't want to be in the business of trying to fix the user's input. But I agree that this might be a common error. I'll make the change to filter ''.

qimingj: Agree that we should not clean up user's text input. Whitespace is special because having it in vocab seems incorrect.

PR #354, opened 2017-04-13
Storage object deletion waits for consistency.

### LINKED ISSUE #344: Storage delete operations are not sync
As pointed out by @craigcitro, `Object.delete` and `Bucket.delete` actually submit delete jobs to the storage service, so they're not exactly synchronous operations, which is different from user expectations.

We should explore either waiting on the jobs to finish as part of the API itself, or rename these operations to add a `_async` suffix.

craigcitro:
Fixes #344.

GCS object deletions are eventually consistent, in that deleting an object in
a globally-available bucket may result in the object still appearing when
listing objects in the bucket:
  https://cloud.google.com/storage/docs/consistency#eventually_consistent_operations

This is a detail, but can lead to confusion for users: someone following the
examples in the tutorial notebooks may see rare-but-regular failures about an
object they just deleted continuing to exist. (We've seen this running the
notebooks as automated tests.) This PR teaches an object to poll the bucket
until it no longer appears. This is controllable by `wait_for_deletion`, and
defaults to **on**.

I've added two tests: first is a mock-based unit test, which works; the second
is intended as a first step towards integration tests. In this case, we assume
that the user running the test has ambient credentials that `google.datalab`
will discover.

### DIFF
--- a/google/datalab/storage/_object.py
+++ b/google/datalab/storage/_object.py
@@ -16,6 +16,7 @@ from __future__ import unicode_literals
 from builtins import object
 
 import dateutil.parser
+import time
 
 import google.datalab
 import google.datalab.utils
@@ -24,6 +25,10 @@ from . import _api
 
 # TODO(nikhilko): Read/write operations don't account for larger files, or non-textual content.
 #                 Use streaming reads into a buffer or StringIO or into a file handle.
+
+# In some polling operations, we sleep between API calls to avoid hammering the
+# server. This argument controls how long we sleep between API calls.
+_POLLING_SLEEP = 1
 
 
 class ObjectMetadata(object):
@@ -136,8 +141,12 @@ class Object(object):
     except Exception as e:
       raise e
 
-  def delete(self):
+  def delete(self, wait_for_deletion=True):
     """Deletes this object from its bucket.
+
+    Args:
+      wait_for_deletion: If True, we poll until this object no longer appears in
+          objects.list operations for this bucket before returning.
 
     Raises:
       Exception if there was an error deleting the object.
@@ -147,6 +156,14 @@ class Object(object):
         self._api.objects_delete(self._bucket, self._key)
       except Exception as e:
         raise e
+      if wait_for_deletion:
+        while True:
+          objects = Objects(self._bucket, prefix=self.key, delimiter='/',
+                            context=self._context)
+          if any(o.key == self.key for o in objects):
+            time.sleep(_POLLING_SLEEP)
+            continue
+          break
 
   @property
   def metadata(self):
--- a/google/datalab/storage/_object.py
+++ b/google/datalab/storage/_object.py
@@ -16,6 +16,7 @@ from __future__ import unicode_literals
 from builtins import object
 
 import dateutil.parser
+import logging
 import time
 
 import google.datalab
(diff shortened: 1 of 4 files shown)


craigcitro:
PTAL @yebrahim

google/datalab/storage/_object.py
```diff
@@ -147,6 +156,14 @@ def delete(self):
         self._api.objects_delete(self._bucket, self._key)
       except Exception as e:
         raise e
+      if wait_for_deletion:
+        while True:
+          objects = Objects(self._bucket, prefix=self.key, delimiter='/',
+                            context=self._context)
+          if any(o.key == self.key for o in objects):
+            time.sleep(_POLLING_SLEEP)
+            continue
```
yebrahim: Do we want to have a timeout for this or are we sure the API never returns except in the success case?

tests/integration/storage_test.py
```diff
@@ -0,0 +1,49 @@
...
+    if not project_id:
+      # If there's no project ID set, we try asking gcloud. We need to send
+      # stderr to /dev/null, otherwise we get some spurious output.
+      gcloud_output = subprocess.check_output([
+          'gcloud', 'config', 'get-value', 'project'], stderr=subprocess.PIPE)
+      project_id = gcloud_output.strip()
+    logging.info('Using project ID: %s', project_id)
+
+    context = google.datalab.Context.default()
+    context.set_project_id(project_id)
```
yebrahim: Why don't we just use `google.datalab.Context.default()` as is? It should pick up the project value from gcloud, just like what the lines 21-31 do.

yebrahim: We should separately have a unit test that makes sure setting a project on gcloud gets read by the `Context.default()` object, so using it directly here should be fine.

google/datalab/storage/_object.py
craigcitro: Ah, good call. Done.

tests/integration/storage_test.py
craigcitro: Oh, nice! I didn't realize that at all.

Fixed.

craigcitro:
I'm happy to add polling on create operations -- my feeling was that we should let this bake, and if it seems to work well, actually add it for *all* eventually consistent operations in one go. Should I file a separate bug to that effect?

tests/integration/storage_test.py
```diff
@@ -0,0 +1,39 @@
...
+
+class StorageTest(unittest.TestCase):
+
+  def setUp(self):
+    context = google.datalab.Context.default()
+    logging.info('Using project: %s', context.project_id)
+    self._context = context
+
+    suffix = ''.join(random.choice(string.letters) for _ in range(8))
+    self._test_bucket_name = '-'.join((project_id, suffix)).lower()
```
yebrahim: `context.project_id`

yebrahim:
Sweet! Yes I think a follow-up issue on eventual consistency should be enough for now.

craigcitro:
https://github.com/googledatalab/pydatalab/issues/355 is the follow-on.

tests/integration/storage_test.py
craigcitro: Done.

PR #359, opened 2017-04-21
One fix in Image package and one fix in google.datalab.ml

qimingj:
Fix an issue in image classification package: trainer incorrectly depend on Dataflow. The fix moves gcsio from Dataflow impl to Tensorflow. Also restructured the importing a little bit so that when loading the package, "import apache_beam" is not hit until it is really used.

Fix an issue in google.datalab.ml.Summary and google.datalab.FeatureSliceView by making it python3 compatible in the correct way.

### DIFF
--- a/google/datalab/ml/_feature_slice_view.py
+++ b/google/datalab/ml/_feature_slice_view.py
@@ -58,7 +58,7 @@ class FeatureSliceView(object):
     """
     import IPython
 
-    if ((sys.version_info.major > 2 and isinstance(data, str)) or 
+    if ((sys.version_info.major > 2 and isinstance(data, str)) or
        (sys.version_info.major <= 2 and isinstance(data, basestring))):
       data = bq.Query(data)
 
--- a/google/datalab/ml/_feature_slice_view.py
+++ b/google/datalab/ml/_feature_slice_view.py
@@ -58,10 +58,8 @@ class FeatureSliceView(object):
     """
     import IPython
 
-    if sys.version_info.major > 2:
-      basestring = (str, bytes)  # for python 3 compatibility
-
-    if isinstance(data, basestring):
+    if ((sys.version_info.major > 2 and isinstance(data, str)) or 
+       (sys.version_info.major <= 2 and isinstance(data, basestring))):
       data = bq.Query(data)
 
     if isinstance(data, bq.Query):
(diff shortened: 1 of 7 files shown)


google/datalab/ml/_feature_slice_view.py
```diff
@@ -58,10 +58,8 @@ def plot(self, data):
     """
     import IPython
 
-    if sys.version_info.major > 2:
-      basestring = (str, bytes)  # for python 3 compatibility
-
-    if isinstance(data, basestring):
+    if ((sys.version_info.major > 2 and isinstance(data, str)) or 
+       (sys.version_info.major <= 2 and isinstance(data, basestring))):
```
brandondutra: it might be cleaner to use python's six package and do

if isinstance(data, six.string_types)

google/datalab/ml/_summary.py
```diff
@@ -95,10 +95,9 @@ def get_events(self, event_names):
           events (i.e. 'loss' under trains_set/, and 'loss' under eval_set/.)
     """
 
-    if sys.version_info.major > 2:
-      basestring = (str, bytes)  # for python 3 compatibility
-
-    event_names = [event_names] if isinstance(event_names, basestring) else event_names
+    if ((sys.version_info.major > 2 and isinstance(event_names, str)) or
+        (sys.version_info.major <= 2 and isinstance(event_names, basestring))):
```
brandondutra: same. Also, six has bools for python 2 and 3. So you could do "if six.PY2"

solutionbox/image_classification/mltoolbox/image/classification/_predictor.py
```diff
@@ -95,7 +95,9 @@ class LoadImagesDoFn(beam.DoFn):
   """A DoFn that reads image from url."""
 
   def process(self, element):
-    with _util.open_local_or_gcs(element['image_url'], 'r') as ff:
+    from tensorflow.python.lib.io import file_io as tf_file_io
```
brandondutra: why local imports again?

qimingj: There is no global "from tensorflow.python.lib.io import file_io as tf_file_io". Do local import in case Dataflow does not pickle the main session, just to be safe.

google/datalab/ml/_summary.py
qimingj: Good idea. I am a bit hesitate to introduce a dependency on six now. Will open an issue for that.

google/datalab/ml/_feature_slice_view.py
qimingj: Good idea. I am a bit hesitate to introduce a dependency on six now. Will open an issue for that.

PR #367, opened 2017-05-03
Fix Pydatalab incompatibilities between Py2/3

chmeyers:
These all have to do with bytes/string differences.


chmeyers:
This will fix at least one of the things seen in #340

PR #368, opened 2017-05-03
added analyze + test file for the new SD package

brandondutra:
* The new SD package does not really have a name yet, so for now I'll put the files in /solutionbox/code_free_ml. Open to other names/locations.
* Test file included but not part of the Travis test build for now. Adding it would require dealing with installing TFT. To run the test locally, datalab and tft 0.1.7 needs to be installed.
* The analyze script takes the SD schema file and features/transforms file, run analysis (locally or with bigquery) and produces a TFT saved model file.
* The analysis part is basically a fork of the files from solutionbox/structured_data/mltoolbox/_structured_data/preprocess
* Building the TFT function for our transforms are implemented in this script as we do analysis outside of beam.

Missing transforms: img url to embedding. Should not be hard to do. I need to look into installing tft from head and playing with Bradley's image functions

Future PRs/roadmap:
1. start a 'csv to tf.example' transform beam script. This is mostly already done. Will clean it up and submit it
1. make a trainer that uses the tft output. Started.
1. then go back and add the image transform
1. online/batch prediction
1. tooling support for images
1. datalab magics/wizard support built on top of this

### DIFF
--- a/solutionbox/code_free_ml/analyze_data.py
+++ b/solutionbox/code_free_ml/analyze_data.py
@@ -26,6 +26,7 @@ import pandas as pd
 import six
 import tensorflow as tf
 import tensorflow_transform as tft
+import textwrap
 
 from tensorflow.contrib import lookup
 from tensorflow.python.lib.io import file_io
@@ -59,6 +60,11 @@ NUMERIC_TRANSFORMS = [IDENTITY_TRANSFORM, SCALE_TRANSFORM]
 CATEGORICAL_TRANSFORMS = [ONE_HOT_TRANSFORM, EMBEDDING_TRANSFROM]
 TEXT_TRANSFORMS = [BOW_TRANSFORM, TFIDF_TRANSFORM]
 
+# If the features file is missing transforms, apply these.
+DEFAULT_NUMERIC_TRANSFORM = IDENTITY_TRANSFORM
+DEFAULT_CATEGORICAL_TRANSFORM = ONE_HOT_TRANSFORM
+
+# Schema values
 INTEGER_SCHEMA = 'integer'
 FLOAT_SCHEMA = 'float'
 STRING_SCHEMA = 'string'
@@ -69,7 +75,7 @@ def parse_arguments(argv):
   """Parse command line arguments.
 
   Args:
-    argv: list of command line arguments, includeing programe name.
+    argv: list of command line arguments, including program name.
 
   Returns:
     An argparse Namespace object.
@@ -78,7 +84,67 @@ def parse_arguments(argv):
     ValueError: for bad parameters
   """
   parser = argparse.ArgumentParser(
-      description='Runs analysis on structured data.')
+      formatter_class=argparse.RawDescriptionHelpFormatter,
+      description=textwrap.dedent("""\
+          Runs analysis on structured data and produces auxiliary files for
+          training. The output files can also be used by the Transform step
+          to materialize TF.Examples files, which for some problems can speed up
+          training.
+
+          Description of input files
+          --------------------------
+
+          1) If using csv files, the --csv-schema-file must be the file path to
+             a schema file. The format of this file must be a valid BigQuery
+             schema file, which is a JSON file containing a list of dicts.
+             Consider the example schema file below:
+
+             [
+                {"name": "column_name_1", "type": "integer"},
+                {"name": "column_name_2", "type": "float"},
+                {"name": "column_name_3", "type": "string"},
+                {"name": "column_name_4", "type": "string"},
+             ]
+
+             Note that the column names in the csv file much match the order
+             in the schema list. Also, we only support three BigQuery types (
+             integer, float, and string).
+
+             If instead of csv files, --bigquery-table is used, the schema file
+             does not have to be pass in as this program will extract it from
+             the table directly.
+
+          2) --features-file is a file path to a file describing the
+             transformations. Below is an example features file:
+
+             {
+                "column_name_1": {"transform": "scale"},
+                "column_name_3": {"transform": "target"},
+                "column_name_2": {"transform": "one_hot"},
+                "column_name_4": {"transform": "key"},
+             }
+
+             The format is of the dict is `name`: `transform-dict` where the
+             `name` must be a column name from the schema file. A list of
+             supported `transform-dict`s is below:
+
+             {"transform": "identity"}: does nothing (for numerical columns).
+             {"transform": "scale", "value": x}: scale a numerical column to
+                [-a, a]. If value is missing, x defaults to 1.
+             {"transform": "one_hot"}: makes a one-hot encoding of a string
+                column.
+             {"transform": "embedding", "embedding_dim": d}: makes an embedding
+                of a string column.
+             {"transform": "bag_of_words"}: bag of words transform for text
+                columns.
+             {"transform": "tfidf"}: TFIDF transform for text columns'
+             {"transform": "target"}: denotes what column is the target. If the
+                schema type of this column is string, a one_hot encoding is
+                automatically applied. If type is numerical, a identity transform
+                is automatically applied.
+             {"transform": "key"}: column contains metadata-like information
+                and is not included in the model.
+  """))
   parser.add_argument('--cloud',
                       action='store_true',
                       help='Analysis will use cloud services.')
@@ -96,18 +162,18 @@ def parse_arguments(argv):
   parser.add_argument('--csv-schema-file',
                       type=str,
                       required=False,
-                      help=('BigQuery json schema file'))
+                      help=('BigQuery json schema file path'))
 
   # If using bigquery table
   parser.add_argument('--bigquery-table',
                       type=str,
                       required=False,
-                      help=('project:dataset.table_name'))
+                      help=('project.dataset.table_name'))
 
   parser.add_argument('--features-file',
                       type=str,
                       required=True,
-                      help='File listing transforms to perform.')
+                      help='Features file path')
 
   args = parser.parse_args(args=argv[1:])
 
@@ -136,6 +202,21 @@ def parse_arguments(argv):
 # ------------------------------------------------------------------------------
 
 def make_scale_tito(min_x_value, max_x_value, output_min, output_max):
+  """Scale a column to [output_min, output_max].
+
+  Assumes the columns's range is [min_x_value, max_x_value]. If this is not
+  true at training or prediction time, the output value of this scale could be
+  outside the range [output_min, output_max].
+
+  Raises:
+    ValueError: if min_x_value = max_x_value, as the column is constant.
+  """
+
+  if round(min_x_value - max_x_value, 7) == 0:
+    # There is something wrong with the data.
+    # Why round to 7 places? It's the same as unittest's assertAlmostEqual.
+    raise ValueError('In make_scale_tito, min_x_value == max_x_value')
+
   def _scale(x):
     min_x_valuef = tf.to_float(min_x_value)
     max_x_valuef = tf.to_float(max_x_value)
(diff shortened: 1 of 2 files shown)


brandondutra:
ping

qimingj:
Reviewing... Sorry for missing it.

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -0,0 +1,823 @@
...
+                      required=False,
+                      help=('BigQuery json schema file'))
+
+  # If using bigquery table
+  parser.add_argument('--bigquery-table',
+                      type=str,
+                      required=False,
+                      help=('project:dataset.table_name'))
+
+  parser.add_argument('--features-file',
```
qimingj: 'transforms-file'?

qimingj: How do we document the file format?

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -0,0 +1,823 @@
...
+
+  # Write the stats file.
+  numerical_results.update(vocab_sizes)
+  stats = {'column_stats': numerical_results, 'num_examples': num_examples}
+  file_io.write_string_to_file(
+      os.path.join(args.output_dir, STATS_FILE),
+      json.dumps(stats, indent=2, separators=(',', ': ')))
+
+
+def check_schema_transform_match(schema, features):
```
qimingj: Should we call it "transforms"?

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -0,0 +1,823 @@
...
+        raise ValueError(
+            'Transform %s not supported by schema %s' % (transform, col_type))
+    elif col_type == STRING_SCHEMA:
+      if transform not in CATEGORICAL_TRANSFORMS + TEXT_TRANSFORMS:
+        raise ValueError(
+            'Transform %s not supported by schema %s' % (transform, col_type))
+    else:
+      raise ValueError('Unsupported schema type %s' % col_type)
+
+  if num_key_transforms != 1 or num_target_transforms != 1:
```
qimingj: As discussed key seems optional and we can allow multiple keys. I guess making it optional might involves more code changes but allowing multiple keys seems not a big change?

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -0,0 +1,823 @@
...
+
+def expand_defaults(schema, features):
+  """Add to features any default transformations.
+
+  Not every column in the schema has an explicit feature transformation listed
+  in the featurs file. For these columns, add a default transformation based on
+  the schema's type. The features dict is modified by this function call.
+
+  Args:
+    schema: schema file
```
qimingj: Not file, not an object. Same for features (should we call it transforms)?

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -0,0 +1,823 @@
...
+    features: features file
+
+  Raises:
+    ValueError: if transform cannot be applied given schema type.
+  """
+
+  schema_names = [x['name'] for x in schema]
+
+  for source_column in six.iterkeys(features):
```
qimingj: "if any(x not in schema_names for x in six.iterkeys(features)):" saves a line. Up to you.

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -0,0 +1,823 @@
...
+SCALE_TRANSFORM = 'scale'
+ONE_HOT_TRANSFORM = 'one_hot'
+EMBEDDING_TRANSFROM = 'embedding'
+BOW_TRANSFORM = 'bag_of_words'
+TFIDF_TRANSFORM = 'tfidf'
+KEY_TRANSFORM = 'key'
+TARGET_TRANSFORM = 'target'
+
+# Transform collections
+NUMERIC_TRANSFORMS = [IDENTITY_TRANSFORM, SCALE_TRANSFORM]
```
qimingj: Add a comment here that the first element is the default transform. Or maybe better with a separate definition "DEFAULT_NUMERIC_TRANSFORMS"?

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -0,0 +1,823 @@
...
+  expand_defaults(schema, features)  # features are updated.
+  check_schema_transform_match(schema, features)
+
+  file_io.recursive_create_dir(args.output_dir)
+
+  if args.cloud:
+    run_cloud_analysis(args, schema, features)
+  else:
+    if not os.path.isdir(args.output_dir):
+      os.makedirs(args.output_dir)
```
qimingj: output_dir is already created above.

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -0,0 +1,823 @@
...
+from tensorflow.contrib import lookup
+from tensorflow.python.lib.io import file_io
+from tensorflow_transform import impl_helper
+from tensorflow_transform.tf_metadata import dataset_metadata
+from tensorflow_transform.tf_metadata import dataset_schema
+from tensorflow_transform.tf_metadata import metadata_io
+
+# Files
+SCHEMA_FILE = 'schema.json'
+FEATURES_FILE = 'features.json'
```
qimingj: Would it be better to call it "TRANSFORMS_FILE"?

solutionbox/code_free_ml/test_analyze_data.py
```diff
@@ -0,0 +1,649 @@
...
+    session = tf.Session(graph=g)
+    with g.as_default():
+      inputs, outputs = impl_helper.load_transform_fn_def(model_path)
+      session.run(tf.tables_initializer())
+      feed = {inputs[key]: value for key, value in six.iteritems(predict_data)}
+      result = session.run(outputs, feed_dict=feed)
+
+    return result
+
+  def test_make_transform_graph_numerics(self):
```
qimingj: Add a test where input from GCS and output to GCS?

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -0,0 +1,823 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
```
brandondutra: Fully agree, but I think we should keep it as one big file for now until we figure out how this will ship/package.

For example, if we want users to call this on the command line like a script, then I think one file is easier.

solutionbox/code_free_ml/analyze_data.py
brandondutra: I originally called this file transforms (and is still called transforms internally) in the original SD package. However, Nikhil wanted this file called feature in the public api. So the current SD package names it a features file. I don't have strong opinions on this name, but we should keep the name to be comparable with SD.

solutionbox/code_free_ml/analyze_data.py
qimingj: OK. Makes sense.

solutionbox/code_free_ml/analyze_data.py
qimingj: OK. Compatible with current SD is good.

solutionbox/code_free_ml/test_analyze_data.py
brandondutra: added test_make_transform_graph_numerics_gcs

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -78,7 +84,67 @@ def parse_arguments(argv):
...
+                {"name": "column_name_3", "type": "string"},
+                {"name": "column_name_4", "type": "string"},
+             ]
+
+             Note that the column names in the csv file much match the order
+             in the schema list. Also, we only support three BigQuery types (
+             integer, float, and string).
+
+             If instead of csv files, --bigquery-table is used, the schema file
+             does not have to be pass in as this program will extract it from
```
qimingj: "the schema file does not have to be pass in" --> "the schema file is not needed".

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -78,7 +84,67 @@ def parse_arguments(argv):
...
+             transformations. Below is an example features file:
+
+             {
+                "column_name_1": {"transform": "scale"},
+                "column_name_3": {"transform": "target"},
+                "column_name_2": {"transform": "one_hot"},
+                "column_name_4": {"transform": "key"},
+             }
+
+             The format is of the dict is `name`: `transform-dict` where the
```
qimingj: extra "is".

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -78,7 +84,67 @@ def parse_arguments(argv):
...
+
+             {"transform": "identity"}: does nothing (for numerical columns).
+             {"transform": "scale", "value": x}: scale a numerical column to
+                [-a, a]. If value is missing, x defaults to 1.
+             {"transform": "one_hot"}: makes a one-hot encoding of a string
+                column.
+             {"transform": "embedding", "embedding_dim": d}: makes an embedding
+                of a string column.
+             {"transform": "bag_of_words"}: bag of words transform for text
+                columns.
```
qimingj: Text columns are no different from string columns from their schema. Should we call them string columns?

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -78,7 +84,67 @@ def parse_arguments(argv):
...
+             {"transform": "tfidf"}: TFIDF transform for text columns'
```
qimingj: ' -> .

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -394,34 +497,37 @@ def preprocessing_fn(inputs):
...
+
+  Returns:
+    TFT Schema object.
   """
   result = {}
 
   # stats file us used to get default values.
   stats = {}
-  if file_io.file_exists(os.path.join(args.output_dir, STATS_FILE)):
+  if file_io.file_exists(os.path.join(output_dir, STATS_FILE)):
```
qimingj: The stats file should always exist according to your impl above. So remove the if?

solutionbox/code_free_ml/analyze_data.py
```diff
@@ -394,34 +497,37 @@ def preprocessing_fn(inputs):
...
       else:
         raise ValueError('unknown transform %s' % transform_name)
     return result
 
   return preprocessing_fn
 
 
-def make_tft_input_schema(schema, args):
-  """Make a tft-stype schema file.
+def make_tft_input_schema(schema, output_dir):
```
qimingj: Nit: calling it outpur_dir is a big misleading as the function does not produce any outputs. Maybe pass in stats_file?

solutionbox/code_free_ml/analyze_data.py
brandondutra: done

solutionbox/code_free_ml/analyze_data.py
brandondutra: done

solutionbox/code_free_ml/analyze_data.py
brandondutra: done. Also added an example of what are the tokens in a string.

solutionbox/code_free_ml/analyze_data.py
brandondutra: done

solutionbox/code_free_ml/analyze_data.py
brandondutra: done

solutionbox/code_free_ml/analyze_data.py
brandondutra: done

brandondutra:
need new approval

PR #370, opened 2017-05-09
transform beam script for the new SD package

brandondutra:
Beam script to read data from csv or bigquery, and apply the tft transformation graph. Output is always tf.example files.

Supports image columns. For images, the file is read and converted to web safe base64 string before passing it to tft.

Also includes local test file. The test file is not part of pydatalab's automatic test framework.

### DIFF
--- a/solutionbox/code_free_ml/transform_raw_data.py
+++ b/solutionbox/code_free_ml/transform_raw_data.py
@@ -1,0 +1,253 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Flake8 cannot disable a warning for the file. Flake8 does not like beam code
+# and reports many 'W503 line break before binary operator' errors. So turn off
+# flake8 for this file.
+# flake8: noqa
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import argparse
+import datetime
+import json
+import logging
+import os
+import random
+import sys
+import apache_beam as beam
+from apache_beam.metrics import Metrics
+from PIL import Image
+import six
+from tensorflow.python.lib.io import file_io
+from tensorflow_transform import coders
+from tensorflow_transform.beam import impl as tft
+from tensorflow_transform.beam import tft_beam_io
+from tensorflow_transform.tf_metadata import metadata_io
+
+
+img_error_count = Metrics.counter('main', 'ImgErrorCount')
+
+# Files
+SCHEMA_FILE = 'schema.json'
+FEATURES_FILE = 'features.json'
+
+TRANSFORMED_METADATA_DIR = 'transformed_metadata'
+RAW_METADATA_DIR = 'raw_metadata'
+TRANSFORM_FN_DIR = 'transform_fn'
+
+# Individual transforms
+TARGET_TRANSFORM = 'target'
+IMAGE_URL_TO_VEC_TRANSFORM = 'img_url_to_vec'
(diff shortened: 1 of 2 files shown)


solutionbox/code_free_ml/transform_raw_data.py
```diff
@@ -0,0 +1,253 @@
...
+
+def parse_arguments(argv):
+  """Parse command line arguments.
+  Args:
+    argv: list of command line arguments including program name.
+  Returns:
+    The parsed arguments as returned by argparse.ArgumentParser.
+  """
+  parser = argparse.ArgumentParser(
+      description='Runs Preprocessing on the Criteo model data.')
```
qimingj: Criteo?

solutionbox/code_free_ml/transform_raw_data.py
```diff
@@ -0,0 +1,253 @@
...
+
+  parser.add_argument(
+      '--project-id',
+      help='The project to which the job will be submitted.')
```
qimingj: Add "Only needed if cloud".

solutionbox/code_free_ml/transform_raw_data.py
```diff
@@ -0,0 +1,253 @@
...
+      action='store_true',
+      help='Run preprocessing on the cloud.')
+  parser.add_argument(
+      '--job-name',
+      type=str,
+      help='Unique job name if running on the cloud.')
+
+  parser.add_argument(
+      '--csv-file-pattern',
+      required=False,
```
qimingj: I would avoid mentioning tf.example, since it is internal impl that users don't have to know. Maybe just "CSV data to transform".

solutionbox/code_free_ml/transform_raw_data.py
```diff
@@ -0,0 +1,253 @@
...
+    raw_data = (
+        pipeline
+        | 'ReadBiqQueryData'
+        >> beam.io.Read(beam.io.BigQuerySource(query=query,
+                                               use_standard_sql=True)))
+
+  raw_data = (
+      raw_data
+      | 'PreprocessTransferredLearningTransformations'
+      >> beam.Map(prepare_image_transforms, features))
```
qimingj: It is not to embeddings yet? Add a TODO to really apply transferred learning?

solutionbox/code_free_ml/transform_raw_data.py
```diff
@@ -0,0 +1,253 @@
...
+  options = {
+      'job_name': args.job_name,
+      'temp_location':
+          os.path.join(args.output_dir, 'tmp'),
+      'project':
+          args.project_id,
+  }
+  pipeline_options = beam.pipeline.PipelineOptions(flags=[], **options)
+
+  temp_dir = os.path.join(args.output_dir, 'tmp')
```
qimingj: This line can be moved above so 'options' can use temp_dir.

solutionbox/code_free_ml/transform_raw_data.py
```diff
@@ -0,0 +1,253 @@
...
+import sys
+import apache_beam as beam
+from apache_beam.metrics import Metrics
+import six
+from tensorflow.python.lib.io import file_io
+from tensorflow_transform import coders
+from tensorflow_transform.beam import impl as tft
+from tensorflow_transform.beam import tft_beam_io
+from tensorflow_transform.tf_metadata import metadata_io
+
```
qimingj: Add some quick comments about why transform?

1. Transform is required if a) there are img_url_to_vec transformation. b) data is in BigQuery.
2. Transform is desired if data transformation is expensive, such as tf-idf is used on text.

solutionbox/code_free_ml/transform_raw_data.py
qimingj: project-id is only needed if transform happens in cloud, or the data is in BigQuery.

solutionbox/code_free_ml/transform_raw_data.py
brandondutra: done

solutionbox/code_free_ml/transform_raw_data.py
brandondutra: done. I think the real work will be done in tft.TransformDataset() once analyze_data.py makes a saved_mode file that does the image-string to embeddings transform.

solutionbox/code_free_ml/transform_raw_data.py
brandondutra: done

PR #372, opened 2017-05-10
make code_free_ml package work with dataflow

brandondutra:
Restructured the current code to be a mltoolbox sub-package (called code_free_ml, but again, we can pick a better name latter)

The code_free_ml package installs the analysis and transform scripts as data files, so it is impossible to access them via the import line! If this package install magics, the code could be placed in /code_free_ml/mltoolbox/code_free_ml

This is a smaller change than what GH says as transform_raw_data.py and test_transform_raw_data.py are almost unchanged. I think I did a mv instead of a git mv the first time, and this missed things up.

Files:
* /code_free_ml/mltoolbox/code_free_ml/data/setup.py datalab and the trainer will use this to install the dependent packages on workers.
* /code_free_ml/mltoolbox/code_free_ml/data/transform_raw_data.py: unchanged except for main()
* /code_free_ml/setup.py: makes code_free_ml pip installable.
* /code_free_ml/test_mltoolbox/test_transform_raw_data.py: unchanged except for the addition of a global CODE_PATH variable and its usage.


brandondutra:
proof you cannot access the script files from the import line:

The script

```
import mltoolbox.code_free_ml as ml
from pkg_resources import resource_filename, resource_listdir

print(dir(ml))
print(resource_listdir('mltoolbox.code_free_ml', 'data'))
print(resource_filename('mltoolbox.code_free_ml', 'analyze_data.py'))
```

prints
```

['__builtins__', '__doc__', '__file__', '__name__', '__package__', '__path__']
['__init__.py', 'transform_raw_data.py', 'analyze_data.py', 'setup.py']
/usr/local/google/home/brandondutra/miniconda2/envs/mltb/lib/python2.7/site-packages/mltoolbox/code_free_ml/analyze_data.py

```

brandondutra:
is this importable? Yes and no.

Does not work:
import mltoolbox.code_free_ml.data as d
dir(d)
['__builtins__', '__doc__', '__file__', '__name__', '__package__', '__path__']

This works:
python -m mltoolbox.code_free_ml.data.analyze_data args

Kinda works:
import mltoolbox.code_free_ml.data.analyze_data as a
 : dir(a)
[ all functions]
So you could call a.main(args). 

What if I remove the __init__.py file in the /data folder?
import mltoolbox.code_free_ml.data.analyze_data as a
ImportError: No module named data.analyze_data

Does python -m mltoolbox.code_free_ml.data.analyze_data work without the __init__ file? No.

PR #373, opened 2017-05-10
UDF: STRUCTS as @param and @return values allowed

hhagblom:
Manual testing performed on a live notebook after
hot-patching the code in a running instance on 
Google Compute Engine.

All the test cases are running and producing real results with a query.

With test-cases
// @param x   ARRAY<STRUCT<index INT64, value STRING>>
// @returns   ARRAY<STRING>

Without leading and trailing spaces around type
// @param x ARRAY<STRUCT<index INT64, value STRING>>
// @returns ARRAY<STRUCT<value STRING>>

With leading and trailing spaces around type
// @param x   ARRAY<STRUCT<index INT64, value STRING>>
// @returns   ARRAY<STRUCT<value STRING>>


hhagblom:
I signed it!

Using my work-email account though. Wasn't sure how to change it
hanspeter.hagblom@jayway.com

yebrahim:
Changes look good to me, but we need to add a regression test to cover this new case. @hhagblom can you take care of this?

hhagblom:
Hi,

I'll look into it. Can't promise anything this week though.

yebrahim:
I think I'll pick this up then, since I'm already in the middle of improving our test coverage. Thanks for the fix @hhagblom.

PR #375, opened 2017-05-12
Add context unit tests

yebrahim:
Also fix a small `Context()` bug where an empty config object results in using the default config.


yebrahim:
The tests are mirroring the google/datalab directory structure, where `_context.py` and `_job.py` are module-level classes.

PR #376, opened 2017-05-12
Bigquery unit tests and fixes

yebrahim:
Test coverage of this file is very bad (currently at about 44%) even though there's a lot of critical logic here not being tested. I'll try to focus more on it.

These tests uncovered two small bugs with query parameters data types and verbose mode.

### DIFF
--- a/tests/kernel/bigquery_tests.py
+++ b/tests/kernel/bigquery_tests.py
@@ -15,6 +15,12 @@ from __future__ import unicode_literals
 import mock
 from oauth2client.client import AccessTokenCredentials
 import unittest
+import json
+import pandas
+try:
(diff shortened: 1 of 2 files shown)


tests/kernel/bigquery_tests.py
```diff
@@ -15,6 +15,12 @@
 import mock
 from oauth2client.client import AccessTokenCredentials
 import unittest
+import json
+import pandas
+try:
+  from StringIO import StringIO
```
chmeyers: I assume these tests also pass in Python 3?

yebrahim: The `except` part below is for Python 3.

PR #379, opened 2017-05-12
More _bigquery module unit tests

yebrahim:
Removed one unused method, and added unit tests for udf, datasource, query execute, and query dryrun cell magics.

### DIFF
--- a/tests/kernel/bigquery_tests.py
+++ b/tests/kernel/bigquery_tests.py
@@ -49,6 +49,36 @@ class TestCases(unittest.TestCase):
   @mock.patch('google.datalab.Context.default')
   def test_udf_cell(self, mock_default_context, mock_notebook_environment):
     env = {}
+    mock_default_context.return_value = TestCases._create_context()
+    mock_notebook_environment.return_value = env
+
+    with self.assertRaises(Exception):
+      google.datalab.bigquery.commands._bigquery._udf_cell({'name': None, 'language': 'js'}, '')
+
+    # no return type
+    cell_body = """
+    // @param word STRING
+    // @param corpus STRING
+    re = new RegExp(word, 'g');
+    return corpus.match(re || []).length;
+    """
+    with self.assertRaises(Exception):
+      google.datalab.bigquery.commands._bigquery._udf_cell({'name': 'count_occurrences',
+                                                            'language': 'js'}, cell_body)
+
+    # too many return statements
+    cell_body = """
+    // @param word STRING
+    // @param corpus STRING
+    // @returns INTEGER
+    // @returns STRING
+    re = new RegExp(word, 'g');
+    return corpus.match(re || []).length;
+    """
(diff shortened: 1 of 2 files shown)


tests/kernel/bigquery_tests.py
```diff
@@ -49,15 +49,43 @@ class TestCases(unittest.TestCase):
   @mock.patch('google.datalab.Context.default')
   def test_udf_cell(self, mock_default_context, mock_notebook_environment):
     env = {}
+    mock_default_context.return_value = TestCases._create_context()
+    mock_notebook_environment.return_value = env
+
+    with self.assertRaises(Exception):
```
jimmc: Can you be more specific with the expected Exception here (and below) to make sure you are not catching the wrong exception?

tests/kernel/bigquery_tests.py
```diff
@@ -49,15 +49,43 @@ class TestCases(unittest.TestCase):
...
+      google.datalab.bigquery.commands._bigquery._udf_cell({'name': None, 'language': 'js'}, '')
```
jimmc: Is this testing for no name, or an empty body? Should there be two tests?

tests/kernel/bigquery_tests.py
yebrahim: The raised exception is a generic `Exception` object though. How important do you think this is in unit tests anyway, where most of the underlying layers are mocked out, and only the one function's logic should be tested?

tests/kernel/bigquery_tests.py
yebrahim: This was for no name, I'll add a test for empty cell body.

yebrahim:
Tests are split between these two because they are two different module. One is testing the actual bigquery functionality (e.g. using the `Query` class), the other is testing the magics we added for the bigquery module (e.g. `%%bq query ...`)

tests/kernel/bigquery_tests.py
jimmc: Definitely better to have this assertRaises than no test, and probably not a big deal that it is not more specific, but I have seen cases with a test like this where a later code changes was causing some other exception to be thrown, and it was not caught by the unit test.

For the future, maybe you could use assertRaisesRegexp and include a piece of the expected error message.

tests/kernel/bigquery_tests.py
```diff
@@ -78,6 +129,14 @@ def test_query_cell(self, mock_default_context, mock_notebook_environment):
 
     # test query creation
     q1_body = 'SELECT * FROM test_table'
+
+    # no query specified
+    with self.assertRaises(Exception):
```
jimmc: Thanks for changing the other calls to assertRaisesRegexp. Maybe change this one also?

yebrahim: This test should actually test for query execution, fixed.

PR #380, opened 2017-05-15
Add "image_to_vec" transform. Implement the analyzer part only. Transform part will come later.

qimingj:


solutionbox/code_free_ml/mltoolbox/code_free_ml/data/analyze_data.py
```diff
@@ -29,6 +29,8 @@
 import textwrap
 
 from tensorflow.contrib import lookup
+from tensorflow.contrib.slim.python.slim.nets.inception_v3 import inception_v3
```
chmeyers: Do we know what the plans from the TF side are for bringing this out of contrib?

qimingj: No idea. If there are any breaking changes from TF we'll have to change our code.

PR #381, opened 2017-05-15
UDF fixes, enable tests

yebrahim:
This PR fixes the following with UDFs:
- Correct defaults for optional parameters
- Fix building UDF sql out of parameter list
- Move check for imports with non-javascript to constructor

I also fixed and enabled UDF tests, and added a few missing tests. Among these I added a regression test for https://github.com/googledatalab/pydatalab/pull/373.

### DIFF
--- a/google/datalab/bigquery/_udf.py
+++ b/google/datalab/bigquery/_udf.py
@@ -13,6 +13,7 @@
 """Google Cloud Platform library - BigQuery UDF Functionality."""
 from __future__ import absolute_import
 from __future__ import unicode_literals
+from past.builtins import basestring
 from builtins import object
 
 
@@ -32,7 +33,7 @@ class UDF(object):
   def code(self):
     return self._code
 
-  def __init__(self, name, code, return_type, params='', language='js', imports=''):
+  def __init__(self, name, code, return_type, params=None, language='js', imports=None):
     """Initializes a UDF object from its pieces.
 
     Args:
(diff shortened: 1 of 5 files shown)


google/datalab/bigquery/_udf.py
```diff
@@ -44,12 +45,21 @@ def __init__(self, name, code, return_type, params='', language='js', imports=''
       language: see list of supported languages in the BigQuery docs
       imports: a list of GCS paths containing further support code.
       """
+    if not isinstance(return_type, basestring):
+      raise TypeError('Argument return_type should be a string. Instead got: ', type(return_type))
+    if params and not isinstance(params, dict):
+      raise TypeError('Argument params should be a dictionary of parameter names and types')
+    if imports and not isinstance(imports, list):
+      raise TypeError('Arguments imports should be a list of GCS string paths')
```
jimmc: Remove s from Arguments

tests/bigquery/udf_tests.py
```diff
@@ -21,44 +20,81 @@
...
-    udf = google.datalab.bigquery.UDF('inputs', [('foo', 'string'), ('bar', 'integer')], 'udf',
-                                      'code')
-    context = TestCases._create_context()
-    query = google.datalab.bigquery.Query(sql, udf=udf, context=context)
-    self.assertEquals('SELECT * FROM (SELECT foo, bar FROM udf(source))', query.sql)
+    # bad return type
+    return_type = ['integer']
+    with self.assertRaises(TypeError):
+      google.datalab.bigquery.UDF('test_udf', code, return_type, params, language, imports)
+    return_type = 'integer'
```
jimmc: I don't think this is a good pattern, where you set some data to a bad value, run a test, then set the data back to a good value for the next test. In particular with this series of tests, because you are asserting there should be a TypeError but with no further details, if you happen to make a mistake setting the value back to a good value, you may not be testing what you expect, as the code may be failing due to your incorrectly set "good" values rather than the expected bad value. This is the reasoning behind the general recommentation to run only a single test in each testing function, so that all state will be reset and you don't have to worry about state from one test contaminating the next test.

One possible approach here would be to write a helper function that creates an object that has all of the fields filled out with good data:
  d = initWithGoodData()  // I'm sure you can come up with a better name
  d.return_type = ['integer']   // bad data for this field
  with self.assertRaisesRegexp(TypeError, 'return_type'):
    google.datalab.bigquery.UDF('test_udf', d.code, d.return_type, d.params, d.language, d.imports);
Then you can put each of these tests into a separate test function.

PR #382, opened 2017-05-17
Make image transform working in transform stage. Update tests.

qimingj:


solutionbox/code_free_ml/test_mltoolbox/test_transform_raw_data.py
```diff
@@ -36,112 +25,79 @@ class TestTransformRawData(unittest.TestCase):
...
+      self.assertEqual(len(image_bytes), 2048)
+      self.assertTrue(any(x != 0.0 for x in image_bytes))
 
     finally:
-      shutil.rmtree(self.output_folder)
+      pass
+      shutil.rmtree(self.working_dir)
 
   def test_local_bigquery_transform(self):
     """Test transfrom locally, but the data comes from bigquery."""
```
brandondutra: Maybe add comment saying this test depends on a data file existing on gcs. Side question, what happens when the image file does not exist? Does beam error or do you get an embedding of all zeros?

qimingj: The original thinking was that since image transformer needs to download a checkpoint from GCS anyway, I would test it using the images from the same bucket. But I agree that we should make the test self contained as much as possible. So I made the change to create images instead of downloading from GCS.

PR #384, opened 2017-05-17
Do not wait for the killed process() in google.datalab.ml.Tensorboard.

qimingj:
I heard in some cases, Tensorboard.stop() hangs although I cannot reproduce the issue.


qimingj:
Tensorboard process is intentionally independent of the notebook kernel process. Even if kernel is reset user should still be able to access previously started Tensorboard. Also even after kernel is reset, users can still call Tensorboard.list() to get the ids and then Tensorboard.stop() to stop them. They are responsible for cleaning up the instances they started.

PR #385, opened 2017-05-18
added new sd trainer and test

brandondutra:
This cl:

analyze_data: tfidf ran into an issue where some shape information was missing. I believe this was related to how we were building the transform graph (via calling 'internal' functions). Now, we build a preprocessing_fn without the use of any analyzers. Then we start a local beam job and run the AnalyzeDataset() beam transform. As there are no analyzers, this does not run over the data source. In fact, the local beam job uses an empty p-collection to construct the transform graph. During my frustrations with getting tfidf working at training time, I changed my tfidf/bow transform to work just like the tft versions (so they do not combine repeated tokens). I do not like this output style, but it is an internal detail users should not care about and we can fix this later.

task.py: 
* training time readers: uses tft function to read tf.example files of transformed data. Uses my own reader functions for csv files of raw data. I put a todo to move my csv reader into tft, then it can be removed.
* serving time readers: we only export a graph that can take json. Using tft serving readers.
* export function: using same custom export function from SD v1.
* this file is ridiculously long, but I want to keep it as one file so users can run it like "python task.py" via the command line.
* for text columns with dnn models, a weighted_sparse_column is added to an embedding_column. At first, I wanted to use an one_hot_column, but I encountered an error with tf learn/layers. I have been emailing Mustafa Ispir (...and David Soergel and Martin Wicke) about this, but there is no current solution.

test_trainer.py: tests {dnn, linear} x {classification, regression} problems. Half start from raw csv data, the other half use transformed tf.example files. Unlike the original sd trainer tests, I no longer check for model loss minimization. Instead, the test focuses on making sure training works and produces a working exported model.

to do list

* does online prediction error if the graph is built for json, and the json has missing values (I hope so)
* add tag transform, alias key to tag
* test this with our customer data
* for dnn models with text columns, use one_hot_columns (figure out why it is broken, need to push TF people harder)
* add image transform support


solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+        checkpoint_path = saver.latest_checkpoint(estimator._model_dir)
+      if not checkpoint_path:
+        raise ValueError("Couldn't find trained model at %s."
+                         % estimator._model_dir)
+
+      export_dir = saved_model_export_utils.get_timestamped_export_dir(
+          export_dir_base)
+
+      with tf_session.Session('') as session:
+        # variables.initialize_local_variables()
```
qimingj: Remove.

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+
+      # Add the extra assets
+      if assets_extra:
+        assets_extra_path = os.path.join(compat.as_bytes(export_dir),
+                                         compat.as_bytes('assets.extra'))
+        for dest_relative, source in assets_extra.items():
+          dest_absolute = os.path.join(compat.as_bytes(assets_extra_path),
+                                       compat.as_bytes(dest_relative))
+          dest_path = os.path.dirname(dest_absolute)
+          gfile.MakeDirs(dest_path)
```
qimingj: gfile supports GCS path right? Maybe using file_io to be consistent?

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+    training_batch_size: An int or scalar `Tensor` specifying the batch size to
+      use.
+    raw_keys: List of string keys giving the order in the csv file.
+    transformed_label_keys
+    convert_scalars_to_vectors: Boolean specifying whether this input_fn should
+      convert scalars into 1-d vectors.  This is necessary if the inputs will be
+      used with `FeatureColumn`s as `FeatureColumn`s cannot accept scalar
+      inputs. Default: True.
+
+    num_epochs
```
qimingj: Arg description. Also no "queue_capacity".

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+      raise ValueError("Key %s does not exist in the schema" % k)
+    if not isinstance(column_schemas[k].representation,
+                      dataset_schema.FixedColumnRepresentation):
+      raise ValueError(("CSV files can only support tensors of fixed size"
+                        "which %s is not.") % k)
+    shape = column_schemas[k].tf_shape().as_list()
+    if shape and shape != [1]:
+      # Column is not a scalar-like value. shape == [] or [1] is ok.
+      raise ValueError(("CSV files can only support features that are scalars "
+                        "having shape []. %s has shape %s")
```
qimingj: "or having shape *", or "or are 1-D tensors".

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+
+  """
+
+  if not raw_keys:
+    raise ValueError("raw_keys must be set.")
+
+  column_schemas = raw_metadata.schema.column_schemas
+
+  # Check for errors.
+  for k in raw_keys:
```
qimingj: Should raw_keys match column_schemas completely? Do we allow raw_keys a subset?

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+  Args:
+    args: the command line args
+
+  Returns:
+    A function that returns a tf.learn experiment object.
+  """
+
+  def get_experiment(output_dir):
+    # Merge schema, input features, and transforms.
+    schema_file_path = os.path.join(args.analysis_output_dir, SCHEMA_FILE)
```
qimingj: Write a small helper function for these?

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+
+    stats_file_path = os.path.join(args.analysis_output_dir, STATS_FILE)
+    if not file_io.file_exists(stats_file_path):
+      raise ValueError('File not found: %s' % stats_file_path)
+    stats = json.loads(file_io.read_file_to_string(stats_file_path).decode())
+
+    target_column_name = get_target_name(features)
+    key_column_name = get_key_name(features)
+    header_names = [col['name'] for col in schema]
+    if not target_column_name or not key_column_name:
```
qimingj: This is for future work item: I think key should be optional (match prediction API).

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+        min_eval_frequency=args.min_eval_frequency,
+        eval_steps=None,
+    )
+
+  # Return a function to create an Experiment.
+  return get_experiment
+
+
+def main(argv=None):
+  """Run a Tensorflow model on the Iris dataset."""
```
qimingj: Iris?

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+
+  return feature_columns
+
+
+def _recursive_copy(src_dir, dest_dir):
+  """Copy the contents of src_dir into the folder dest_dir.
+  Args:
+    src_dir: gsc or local path.
+    dest_dir: gcs or local path.
+  When called, dest_dir should exist.
```
brandondutra: done

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+      outputs[PG_TARGET] = tf.squeeze(input_ops.features[target_name])
+
+    scores = model_fn_ops.predictions['scores']
+    outputs[PG_REGRESSION_PREDICTED_TARGET] = tf.squeeze(scores)
+
+  return outputs
+
+
+# This function is strongly based on
+# tensorflow/contrib/learn/python/learn/estimators/estimator.py:export_savedmodel()
```
brandondutra: done

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+    args: command line args
+    keep_target: If ture, target column is returned in prediction graph. Target
+        column must also exist in input data
+    assets_extra: other fiels to copy to the output folder
+    job_dir: root job folder
+    features: features dict
+    schema: schema list
+  """
+  target_name = get_target_name(features)
+  raw_metadata = raw_metadata = metadata_io.read_metadata(
```
brandondutra: ops

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+
+  def export_fn(estimator, export_dir_base, checkpoint_path=None, eval_result=None):
+    with ops.Graph().as_default() as g:
+      contrib_variables.create_global_step(g)
+
+      input_ops = input_fn_maker.build_default_transforming_serving_input_fn(
+          raw_metadata=raw_metadata,
+          transform_savedmodel_dir=os.path.join(args.analysis_output_dir, TRANSFORM_FN_DIR),
+          raw_label_keys=[target_name],
+          raw_feature_keys=csv_header,
```
brandondutra: correct. This is because it is not used in build_feature_columns().

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
brandondutra: done

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
brandondutra: done

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
brandondutra: thanks

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
brandondutra: No. Yes. TFT style input readers in mappers.py allow for only using a subset of the data.

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
brandondutra: I don't think this is true unless you extend the definition of csv files. How do you encode the 3d vector (1,2,3) (which is a 1-D tensor) as one csv column?

Maybe "data,1 2 3,data" or "data,1|2|3,data"?  If you do this, you just invented a new file format called "csv+"!

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
brandondutra: done

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
brandondutra: fully agree, added todo in make_prediction_output_tensors()

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
brandondutra: ops

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -0,0 +1,849 @@
...
+                            'in seconds'))
+
+  args, remaining_args = parser.parse_known_args(args=argv[1:])
+
+  # All HP parambeters must be unique, so we need to support an unknown number
+  # of --layer_size1=10 --layer_size2=10 ...
+  # Look at remaining_args for layer_size\d+ to get the layer info.
+
+  # Get number of layers
+  pattern = re.compile('layer-size(\d+)')
```
qimingj: Layers that are not input or output layers, are hidden layers. See https://en.wikipedia.org/wiki/Feedforward_neural_network.

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/analyze_data.py
```diff
@@ -245,15 +245,14 @@ def _scale(x):
...
-    x: A `Column` representing a string value or values.
+    x: A `Column` representing a string value.
     vocab: list of strings.
 
   Returns:
-    A `Column` where each string value is mapped to an integer where each unique
-    string value is mapped to a different integer and integers are consecutive
-    and starting from 0.
+    A `Column` where each string value is mapped to an integer representing
+    its index in the vocab. Our of vocab values are mapped to len(vocab).
```
qimingj: Out.

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
qimingj: I see. I misunderstood "having shape []".

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/analyze_data.py
brandondutra: done

PR #388, opened 2017-05-24
Support image column in trainer.

qimingj:
If transform is image_to_vec, then add a hidden layer on top of embeddings. This is to match the inception graph design, and probably the best for our graph too.

Also add base64 encoding/decoding to images to make it work with CSV lines input.

### DIFF
--- a/solutionbox/code_free_ml/test_mltoolbox/test_analyze_data.py
+++ b/solutionbox/code_free_ml/test_mltoolbox/test_analyze_data.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+import base64
 import copy
 import cStringIO
 from PIL import Image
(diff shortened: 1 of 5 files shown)


solutionbox/code_free_ml/test_mltoolbox/test_analyze_data.py
```diff
@@ -715,7 +716,7 @@ def _open_and_encode_image(img_url):
         img = Image.open(f).convert('RGB')
         output = cStringIO.StringIO()
         img.save(output, 'jpeg')
-      return output.getvalue()
+      return base64.urlsafe_b64encode(output.getvalue())
```
brandondutra: so the tft image transform expects this column to be base64, but there is no decode op in analyze_data.py

Where is this decoded to regular bytes?

I think analyze_data and transform_raw_data assume the input tensor is a regular byte string. But the code in the test files show the exported graph assumes the column is a base64 string.

Said again, I thin analyze_data and transform_raw_data need updating to start from a base64 string. This might be wasteful. Another option is to have a custom serving input function that applies tf.decode_base64

qimingj: You are right. I did not include 2 files that I should have (they are in different dir)!

Added.

PR #389, opened 2017-05-24
added a local prediction scrip

brandondutra:
assumes the graph takes csv input. Can output csv or json files.

The test file only tests the local prediction script, not the output of training.


solutionbox/code_free_ml/mltoolbox/code_free_ml/data/local_predict.py
```diff
@@ -0,0 +1,251 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
```
qimingj: Since cloud prediction is done by gcloud, should we name this file predict.py?

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/local_predict.py
```diff
@@ -0,0 +1,251 @@
...
+  # I/O args
+  parser.add_argument('--predict-data',
+                      required=True,
+                      help='Data to run prediction on')
+  parser.add_argument('--trained-model-dir',
+                      required=True,
+                      help='Usually train_output_path/model.')
+  parser.add_argument('--output-location',
+                      required=True,
+                      help=('Location to save output.'))
```
qimingj: nit: No need for ().

qimingj: same for next arg.

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/local_predict.py
```diff
@@ -0,0 +1,251 @@
...
+                      action='store_false',
+                      help='Don\'t shard files')
+  parser.set_defaults(shard_files=True)
+
+  args, _ = parser.parse_known_args(args=argv[1:])
+
+  return args
+
+
+def batch_csv_reader(csv_file, batch_size):
```
qimingj: Can this be:

def batch_csv_reader(csv_file, n):
  with file_io.FileIO(csv_file, 'r') as f:
    args = [f] * n
    return izip_longest(*args)

The only difference is it fills Nones in last batch. Seems simpler but up to you.

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/local_predict.py
```diff
@@ -0,0 +1,251 @@
...
+
+  return schema
+
+
+def main(argv=None):
+  """Runs local batch prediction."""
+  args = parse_arguments(sys.argv if argv is None else argv)
+
+  file_io.recursive_create_dir(args.output_location)
+  with tf.Graph().as_default(), tf.Session() as sess:
```
qimingj: I feel "run_batch_prediction" can include the logic of loading model and run session so all TF stuff is wrapped into one function. What do you think?

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/local_predict.py
```diff
@@ -0,0 +1,251 @@
...
+  """
+  _, csv_tensor_name = input_alias_map.items()[0]
+  raw_result = session.run(fetches=output_alias_map,
+                           feed_dict={csv_tensor_name: batch_csv_strings})
+
+  # Convert from numpy arrays to python lists
+  result = {key: raw_result[key].flatten().tolist() for key in raw_result}
+
+  batch_size = len(batch_csv_strings)
+  json_list = []
```
qimingj: A little bit odd to me to call it "json_list" and "json_result" below. Maybe just "results"?

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/local_predict.py
```diff
@@ -0,0 +1,251 @@
...
+  parser.add_argument('--predict-data',
+                      required=True,
+                      help='Data to run prediction on')
+  parser.add_argument('--trained-model-dir',
+                      required=True,
+                      help='Usually train_output_path/model.')
+  parser.add_argument('--output-location',
+                      required=True,
+                      help=('Location to save output.'))
+  parser.add_argument('--output-format',
```
qimingj: Do we also need to support output-file-name? Otherwise for example in Datalab, I have to look for "predictions-xxxxx-of-xxxxx" files, and the "predictions" keyword need to be documented and cannot be changed. In the case of no sharding, I would expect to use output-file-name without appending 'xxxxx-of-xxxxx'.

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/local_predict.py
brandondutra: I used that name at first, but then I thought it might be confusing, so I named it local_predict

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/local_predict.py
brandondutra: done

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/local_predict.py
brandondutra: done. Now there is a --output-filename. It is used "as is" if no shardding is used, else -xxxxx-of-xxxxx is added. File extension is also added based on output format.

brandondutra:
Not continuing with the rest of these review comments as we'll just use gcloud local predict.  Going to close this PR.

gcloud local predict may not do batching

PR #391, opened 2017-05-25
make keys optional

brandondutra:
also allow many keys, and fixed a bug that the exported graph has the incorrect friendly tensor names if there was one input/output tensor name.


solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/task.py
```diff
@@ -302,10 +302,11 @@ def make_prediction_output_tensors(args, features, input_ops, model_fn_ops,
                                    keep_target):
   """Makes the final prediction output layer."""
   target_name = get_target_name(features)
-  key_name = get_key_name(features)  # TODO(brandondutra): make optional, rename to tag
+  key_names = get_key_names(features)  # TODO(brandondutra): make optional, rename to tag
```
qimingj: Do we still need the comments? I think calling them key is fine.

brandondutra: ok, done

PR #393, opened 2017-05-26
More bigquery tests, html tests

yebrahim:
Also fixed two bugs with the table cell magic.

### DIFF
--- a/google/datalab/bigquery/commands/_bigquery.py
+++ b/google/datalab/bigquery/commands/_bigquery.py
@@ -721,7 +721,7 @@ def _table_cell(args, cell_body):
       else:
         context = google.datalab.Context(args['project'],
                                          google.datalab.Context.default().credentials)
-        datasets = [google.datalab.bigquery.Dataset(args['project'], context)]
+        datasets = [google.datalab.bigquery.Dataset(args['dataset'], context)]
     else:
       default_context = google.datalab.Context.default()
       context = google.datalab.Context(default_context.project_id, default_context.credentials)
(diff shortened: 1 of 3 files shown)


google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -719,12 +719,14 @@ def _table_cell(args, cell_body):
       if args['project'] is None:
         datasets = [google.datalab.bigquery.Dataset(args['dataset'])]
       else:
-        datasets = [google.datalab.bigquery.Dataset((args['project'], args['dataset']))]
+        context = google.datalab.Context(args['project'],
```
jimmc: I don't fully understand this code, but this code branch is only executed when the user has specified both project and dataset, yet it looks like in the updated code you no longer use the dataset arg. If this is correct, it probably warrants a comment. Or maybe even throw an exception if the user specifies both and we are going to ignore the dataset arg.

tests/kernel/html_tests.py
```diff
@@ -33,8 +35,40 @@ def noop_decorator(func):
...
+            {'v': 'val1'},
+            {'v': 'val2'}
+          ]},
+          {'c': [
+            {'v': 'val3'},
+            {'v': 'val4'}
+          ]}
+        ]
+      }, ['col1', 'col2'], 'chartdata')
+    expected_html = ''.join('''\
```
jimmc: I think you don't need the backslash.

google/datalab/bigquery/commands/_bigquery.py
yebrahim: Oh nice catch! It's interesting I also made the same mistake in the test, and wrote it to expect project twice! :)

PR #399, opened 2017-06-01
remove tft usage

brandondutra:
* Transforms are now in trainer/feature_transforms.py
* analysis no longer makes a saved model
* transform_raw_data.py now builds the transform graph in memory and does not write it to disk
* nothing uses tensorflow_transform

this cl is large enough, so let me split the following into future cls:
* make a constants file as many files uses the same constants
* double check everything works on the cloud services
* update the optimizer for linear models
* task.py: schema file in output/model should not have the target column


solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -0,0 +1,713 @@
...
+IMAGE_DEFAULT_STRING = base64.urlsafe_b64encode(_img_buf.getvalue())
+
+
+# ------------------------------------------------------------------------------
+# ------------------------------------------------------------------------------
+# start of transform functions
+# ------------------------------------------------------------------------------
+# ------------------------------------------------------------------------------
+
+def scale(x, min_x_value, max_x_value, output_min, output_max):
```
brandondutra: done

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -0,0 +1,713 @@
...
+      elif transform_name == 'scale':
+        result[name] = scale(
+            inputs[name],
+            min_x_value=stats['column_stats'][name]['min'],
+            max_x_value=stats['column_stats'][name]['max'],
+            output_min=transform.get('value', 1) * (-1),
+            output_max=transform.get('value', 1))
+      elif transform_name in [ONE_HOT_TRANSFORM, EMBEDDING_TRANSFROM,
+                              TFIDF_TRANSFORM, BOW_TRANSFORM]:
+        vocab_str = file_io.read_file_to_string(
```
brandondutra: We need the file in memory as we pass it to string_to_index_table_from_tensor. If we used index_table_from_file, we would not need to read it directly, but we would have to change the format of your vocab files (one line in our vocab file is "token,example_count"). In any event, the vocab has to be in memory at some point to do the lookup, right?

What is your suggestion with file_io.FileIO? Do you want me to loop over the file line-by-line, and split the columns into two lists manually?

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -0,0 +1,713 @@
...
+    is None, then the tensor is a sparse tensor.
+  """
+
+  info = collections.defaultdict(dict)
+
+  for name, transform in six.iteritems(features):
+    transform_name = transform['transform']
+
+    if transform_name == IDENTITY_TRANSFORM:
+      schema_type = [col['type'] for col in schema if col['name'] == name]
```
brandondutra: done

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -0,0 +1,713 @@
...
+      schema_type = schema_type[0].lower()
+      if schema_type == FLOAT_SCHEMA:
+        info[name]['dtype'] = tf.float32
+      elif schema_type == INTEGER_SCHEMA:
+        info[name]['dtype'] = tf.int64
+      else:
+        info[name]['dtype'] = tf.string
```
brandondutra: done, now raise an error

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -0,0 +1,713 @@
...
+    elif transform_name == IMAGE_TRANSFORM:
+      info[name]['dtype'] = tf.float32
+      info[name]['size'] = 2048
+    else:
+      raise ValueError('Unknown transfrom %s' % transform_name)
+
+  return info
+
+
+# TODO(brandondutra): should numerical defaults be the mean or 0?
```
brandondutra: updated the todo

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -0,0 +1,713 @@
...
+                               name='csv_input_placeholder')
+  tensors = tf.decode_csv(placeholder, record_defaults)
+  raw_features = dict(zip(csv_header, tensors))
+
+  transform_fn = make_preprocessing_fn(analysis_path, features, keep_target)
+  transformed_tensors = transform_fn(raw_features)
+
+  transformed_features = {}
+  for k, v in six.iteritems(transformed_tensors):
+    if isinstance(v, tf.Tensor) and v.get_shape().ndims == 1:
```
brandondutra: I copied this from some tft code. But you are right, I don't think so. Tests pass with it removed, so no!

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -0,0 +1,713 @@
...
+      files.extend(file_io.get_matching_files(path))
+
+    filename_queue = tf.train.string_input_producer(
+        files, num_epochs=num_epochs, shuffle=randomize_input)
+
+    csv_id, csv_lines = tf.TextLineReader().read_up_to(filename_queue, training_batch_size)
+
+    queue_capacity = (reader_num_threads + 3) * training_batch_size + min_after_dequeue
+    if randomize_input:
+      batch_csv_id, batch_csv_lines = tf.train.shuffle_batch(
```
brandondutra: Correct, but we still need to pass a list to shuffle_batch and batch, and remove the returned list. So we would need to do

batch_csv_lines, = tf.train.shuffle_batch(tensors=[csv_lines], ...)

but then the ',' looks odd and needs a comment. So instead I'll just replace batch_csv_id with '_'

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -0,0 +1,713 @@
...
+          num_threads=reader_num_threads)
+
+    csv_header, record_defaults = csv_header_and_defaults(features, schema, keep_target=True)
+    parsed_tensors = tf.decode_csv(batch_csv_lines, record_defaults, name='csv_to_tensors')
+    raw_features = dict(zip(csv_header, parsed_tensors))
+
+    transform_fn = make_preprocessing_fn(analysis_output_dir, features, keep_target=True)
+    transformed_tensors = transform_fn(raw_features)
+
+    # Exapnd te dimes of non-sparse tensors. This is needed by tf.learn.
```
brandondutra: thanks

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -0,0 +1,713 @@
...
+    # Remove the target tensor, and return it directly
+    target_name = None
+    for name, transform in six.iteritems(features):
+      if transform['transform'] == TARGET_TRANSFORM:
+        target_name = name
+        break
+    if not target_name or target_name not in transformed_features:
+      raise ValueError('Cannot find target transform in features')
+
+    transformed_labels = transformed_features.pop(target_name)
```
brandondutra: Done. I was using tf.learn lingo, but target is better.

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -0,0 +1,713 @@
...
+        files, num_epochs=num_epochs, shuffle=randomize_input)
+
+    options = tf.python_io.TFRecordOptions(
+        compression_type=tf.python_io.TFRecordCompressionType.GZIP)
+    ex_id, ex_str = tf.TFRecordReader(options=options).read_up_to(
+        filename_queue, training_batch_size)
+
+    queue_capacity = (reader_num_threads + 3) * training_batch_size + min_after_dequeue
+    if randomize_input:
+      batch_ex_id, batch_ex_str = tf.train.shuffle_batch(
```
brandondutra: same as above, batch_ex_id is now _

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -0,0 +1,713 @@
...
+    # Exapnd te dimes of non-sparse tensors. This is needed by tf.learn.
+    transformed_features = {}
+    for k, v in six.iteritems(parsed_tensors):
+      if isinstance(v, tf.Tensor) and v.get_shape().ndims == 1:
+        transformed_features[k] = tf.expand_dims(v, -1)
+      else:
+        transformed_features[k] = v
+
+    # Remove the target tensor, and return it directly
+    target_name = None
```
brandondutra: done. Moved the function to this file.

brandondutra:
running tests now, will push changes soon

brandondutra:
I think the only comment I didn't fix was the one about file_io.FileIO as I'm not sure what your suggestion was.

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
qimingj: We do need to store the file content in memory, but reading into into a string, and then parse it by pd.read_csv, now you have 2 copies of the file content in memory. What I suggested is:

with file_io.FileIO(file_path, 'r') as f:
  vocab_pd = pd.read_csv(f, ...)

So we don't load whole file content into memory once, but read it as a stream.
Eventually, the memory used by vocab_str will be garbage collected anyway even if we read all file content, but doing so increases peak memory usage which seems can be avoided easily.

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -689,25 +673,49 @@ def transformed_training_input_fn():
...
 
   return transformed_training_input_fn
+
+
+def get_target_name(features):
+  for name, transform in six.iteritems(features):
+    if transform['transform'] == TARGET_TRANSFORM:
+      return name
+  return None
+
```
qimingj: nit: one more blank line.

brandondutra: done

PR #400, opened 2017-06-05
small fixes

brandondutra:
* move all constants to the feature_transforms file
* exported graph in /model has a schema file with target column removed
* numerical defaults use the stats file
* linear models now use an optimizer that has L1/L2 terms. I did not use the SDCA optimizer because
  1. I think it is only for (binary) classification problems
  1. requires the target column to be a string, not an int
  1. is in tf.contrib, so one less thing that will break in the future
* For linear models, I use the tf.train.FtrlOptimizer optimizer, which is tf's default.

### DIFF
--- a/solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
+++ b/solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
@@ -472,10 +472,10 @@ def csv_header_and_defaults(features, schema, stats, keep_target):
     # Note that numerical key columns do not have a stats entry, hence the use
     # of get(col['name'], {})
     csv_header.append(col['name'])
-    if col['type'].lower() == 'integer':
+    if col['type'].lower() == INTEGER_SCHEMA:
       dtype = tf.int64
       default = int(stats['column_stats'].get(col['name'], {}).get('mean', 0))
-    elif col['type'].lower() == 'float':
+    elif col['type'].lower() == FLOAT_SCHEMA:
       dtype = tf.float32
       default = float(stats['column_stats'].get(col['name'], {}).get('mean', 0.0))
     else:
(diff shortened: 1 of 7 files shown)


brandondutra:
ping

qimingj:
looking

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -64,6 +64,8 @@
 Image.new('RGB', (16, 16)).save(_img_buf, 'jpeg')
 IMAGE_DEFAULT_STRING = base64.urlsafe_b64encode(_img_buf.getvalue())
 
+IMAGE_BOTTLENECK_TENSOR_SIZE = 2048
```
qimingj: There are two types of constants here: one type is part of the public interface, such as schema, transform. Everything else is internal use only. I wonder if we should move the public ones into a different file, so it is easy to point people to the public constant file if they need help creating schema or transforms. The other advantage is that we can keep track of constants that can cause backward compatibility issues.

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
```diff
@@ -464,23 +465,28 @@ def csv_header_and_defaults(features, schema, keep_target):
     if not keep_target and col['name'] == target_name:
       continue
 
+    # Note that numerical key columns do not have a stats entry, hence the use
+    # of get(col['name'], {})
     csv_header.append(col['name'])
     if col['type'].lower() == 'integer':
-      default, dtype = 0, tf.int64
+      dtype = tf.int64
```
qimingj: Should the "integer" above and "float" below be constants?

brandondutra: They are. After the if-else, they are wrapped in tf.constant

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
brandondutra: Only the individual transform names are exposed to users. We should have documentation/help pages instead of pointing to code in our "code free solution"!

Added comments about what constants are 'safe' to change.

solutionbox/code_free_ml/mltoolbox/code_free_ml/data/trainer/feature_transforms.py
brandondutra: ops, yes. done

PR #401, opened 2017-06-06
Add a local prediction magic (%%ml predict).

qimingj:
- Users need to explicit call import google.datalab.contrib.mltoolbox.commands to use the command.
- Local batch prediction will come later.


google/datalab/contrib/mltoolbox/_local_predict.py
```diff
@@ -0,0 +1,167 @@
...
+
+  if isinstance(data, pd.DataFrame):
+    data = data.T.to_dict().values()
+  elif isinstance(data[0], six.string_types):
+    data = list(csv.DictReader(data, fieldnames=headers))
+
+  images = _download_images(data, img_cols)
+  predict_data = _get_predicton_csv_lines(data, headers, images)
+  display_data = data
+  if show_image:
```
brandondutra: if show_image, the image url is still displayed, and an image column is appended. But because everything is a dict, all order information is lost. So the output could be

image 1, image 1 path, image 2 path, image 3 path, image 3, image 2.

I think you want to force an ordering in the final df you return.

google/datalab/contrib/mltoolbox/commands/_ml.py
```diff
@@ -0,0 +1,110 @@
...
+      prog='%ml',
+      description="""
+Execute MLToolbox operations.
+
+Use "%ml <command> -h" for help on a specific command.
+""")
+
+  predict_parser = parser.subcommand(
+      'predict', help='Predict with local or deployed models. ' +
+      'Prediction data is provided in cell in yaml format of CSV lines. For example, ' +
```
brandondutra: Move the data format example to the --data flag, as the input could be other sources

google/datalab/contrib/mltoolbox/commands/_ml.py
```diff
@@ -0,0 +1,110 @@
...
+
+  return google.datalab.utils.commands.handle_magic_line(line, cell, parser)
+
+
+def _predict(args, cell):
+  headers = args['headers'].split(',')
+  img_cols = args['image_columns'].split(',') if args['image_columns'] else []
+  env = google.datalab.utils.commands.notebook_environment()
+  cell_data = google.datalab.utils.commands.parse_config(cell, env)
+  if 'prediction_data' not in cell:
```
brandondutra: I think you need to be checking --data first, and if it is missing then inspect the cell

google/datalab/contrib/mltoolbox/commands/_ml.py
```diff
@@ -0,0 +1,110 @@
...
+                                             img_cols, args['show_image'])
+
+  def _show_img(img_bytes):
+    return '<img src="data:image/png;base64,' + img_bytes + '" />'
+
+  def _truncate_text(text):
+    return (text[:37] + '...') if isinstance(text, six.string_types) and len(text) > 40 else text
+
+  # Truncate text explicitly here because we will set display.max_colwidth to -1.
+  formatters = {x: _truncate_text for x in df.columns if df[x].dtype == np.object}
```
brandondutra: add comment saying this line applied to text and images, but images will be overridden in the next if statement.

tests/mltoolbox_magic/local_predict_tests.py
```diff
@@ -0,0 +1,169 @@
...
+    if embedding_images:
+      for d in data:
+        with open(d['img_url1'], 'r') as f:
+          d['img_url1'] = base64.urlsafe_b64encode(f.read())
+
+    if missing_values:
+      del data[0]['num1']
+      del data[1]['img_url1']
+
+    csv_lines = []
```
brandondutra: nit: this can be moved inside the if

tests/mltoolbox_magic/local_predict_tests.py
```diff
@@ -0,0 +1,169 @@
...
+
+    model_dir = self._create_model('model1')
+    headers = ['key', 'num1', 'text1', 'img_url1']
+
+    # Test data being list of dict and list of csvlines, with and without missing values,
+    # show or not show images.
+    for missing_values in [True, False]:
+      for csv_data in [True, False]:
+        for show_image in [True, False]:
+          test_data = self._create_test_data(False, missing_values, csv_data)
```
brandondutra: You may want to log what the current parameter value is, as when the test fails it will be harder to debug.

tests/mltoolbox_magic/local_predict_tests.py
```diff
@@ -0,0 +1,169 @@
...
+          df = _local_predict.get_prediction_results(
+              model_dir, test_data, headers, ['img_url1'], show_image)
+          self._validate_results(df, show_image)
+
+    # Test data being dataframes, with and without missing values, and embedded images.
+    for missing_values in [True, False]:
+      test_data = self._create_test_data(True, missing_values, csv_data=False)
+      df_s = pd.DataFrame(test_data).fillna('')
```
brandondutra: Maybe get_prediction_results should handle the missing values.

google/datalab/contrib/mltoolbox/_local_predict.py
qimingj: Good point! It appears that pandas DataFrame created from a list of dict uses alphabetical order of column names, and underscore is after uppercase characters. Given prediction data: ['key', 'num', 'text', 'img_url', 'img_urlA'], it will mess up the order: ['key', 'num', 'text', 'img_url', 'img_urlA', image_urlA_image, img_url_image].

To avoid this issue, also to avoid column name clash, instead of appending "_image" now I append " (image)". Space is ordered before all non-control characters so it should work.

google/datalab/contrib/mltoolbox/commands/_ml.py
qimingj: The example only shows how to provide data in input cell, not by --data so should not be in "--data" section. I changed the help string to make it clearer.

qimingj: Actually, --data is not needed. If I have: "mydata = [...]" defined in a cell, then "--data $mydata" will not work because $mydata will be the string repr of the object. I think we can be consistent: always take prediction data from cell. So if data is an object, then we do:

%%ml predict
prediction_data: $mydata

Of course it works too if specifying data directly:

%%ml predict:
prediction_data:
  - value1,value2...
  - value1,value2...

So I removed "--data".

google/datalab/contrib/mltoolbox/commands/_ml.py
qimingj: --data is removed.

google/datalab/contrib/mltoolbox/commands/_ml.py
qimingj: done.

tests/mltoolbox_magic/local_predict_tests.py
qimingj: done.

tests/mltoolbox_magic/local_predict_tests.py
qimingj: Good idea. Done.

tests/mltoolbox_magic/local_predict_tests.py
qimingj: Can't. Otherwise %%ml predict will need to access stats files, which is not possible in online prediction.

qimingj:
Need approval after fixing a bunch of py3 issues. Thanks.

PR #404, opened 2017-06-09
Add a shell process management helper function, which will be used in mltoolbox analyze, transform, train, etc.

qimingj:


google/datalab/contrib/mltoolbox/_shell_process.py
```diff
@@ -0,0 +1,74 @@
...
+  Args:
+    args: the args of the process to start and monitor.
+    pid_to_wait: the process to wait on. If the process ends, also kill the started process.
+    std_out_filter_fn: a filter function which takes a string content from the stdout of the
+        started process, and returns True if the string should be redirected to console stdout.
+  """
+
+  monitor_process = None
+  try:
+    p = subprocess.Popen(args,
```
brandondutra: I think we need the command string (and shell=True)...something like "cd x&& python -m trainer.task --args" will not work

google/datalab/contrib/mltoolbox/_shell_process.py
```diff
@@ -0,0 +1,74 @@
...
+
+def _wait_and_kill(pid_to_wait, pids_to_kill):
+  """ Wait for a process to finish if it exists, and then kill a list of processes.
+
+  Args:
+    pid_to_wait: the process to wait for.
+    pids_to_kill: a list of processes to kill after the process of pid_to_wait finishes.
+  """
+  if psutil.pid_exists(pid_to_wait):
+    psutil.Process(pid=pid_to_wait).wait()
```
yebrahim: timeout?

yebrahim: and maybe exit without killing those processes in the timeout case?

tests/mltoolbox_magic/shell_process_tests.py
```diff
@@ -0,0 +1,88 @@
...
+    self._logger = logging.getLogger('TestStructuredDataLogger')
+    self._logger.setLevel(logging.DEBUG)
+    if not self._logger.handlers:
+      self._logger.addHandler(logging.StreamHandler(stream=sys.stdout))
+
+  def test_process(self):
+    """ Test starting a process and have it depend on another process.
+    """
+
+    process_to_wait_args = ['python', '-c', 'import time; time.sleep(10)']
```
yebrahim: 10 seconds seems a little excessive, do we need all of this waiting in this test?

tests/mltoolbox_magic/shell_process_tests.py
```diff
@@ -0,0 +1,88 @@
...
+    old_stdout = sys.stdout
+    buf = six.StringIO()
+    sys.stdout = buf
+    t = threading.Thread(target=_shell_process.run_and_monitor,
+                         args=(worker_process_args,
+                               process_to_wait.pid, lambda x: x.startswith('include')))
+    t.start()
+    time.sleep(1)
+    process_to_monitor = None
+    for proc in psutil.process_iter():
```
brandondutra: maybe add comment saying something like the thread makes a new process, but we cannot know it's id, so we search for it by it's command line.

tests/mltoolbox_magic/shell_process_tests.py
brandondutra: I don't think it matters. It should be killed before it is done, assuming the test gets to process_to_wait.kill(). Maybe add a comment saying this

tests/mltoolbox_magic/shell_process_tests.py
```diff
@@ -0,0 +1,88 @@
...
+    self.assertIsNone(process_to_wait.poll())
+    self._logger.debug('TestProcess: Started a process %d which will be waited on.' %
+                       process_to_wait.pid)
+
+    worker_process_args = [
+        'python',
+        '-c',
+        'import time;import sys;print(\'exclude message\');time.sleep(1);' +
+        'print(\'include message1\');sys.stdout.flush();time.sleep(3);print(\'include message2\')'
+    ]
```
brandondutra: could the sleep(1) make this test flaky because later you wait 1 sec before killing the process, so maybe message1 will not get printed?

tests/mltoolbox_magic/shell_process_tests.py
```diff
@@ -0,0 +1,88 @@
...
+class TestShellProcess(unittest.TestCase):
+  """Tests for process managements used in MLToolbox magics."""
+
+  def setUp(self):
+    self._logger = logging.getLogger('TestStructuredDataLogger')
+    self._logger.setLevel(logging.DEBUG)
+    if not self._logger.handlers:
+      self._logger.addHandler(logging.StreamHandler(stream=sys.stdout))
+
+  def test_process(self):
```
brandondutra: This test depends on events happening within time intervals. I would like to see a comment/documentation outlining the timeline of events . Something like

step 1: process_to_wait is created with 10 seconds to live. It is assumed this test will finish before then.
step 2: monitor process is created, which then makes a worker process. This test sleeps for 1 sec
step 3: worker process prints "exclude message"
step 4: worker sleeps for 1 sec
step 5: worker prints include message 1
step 6: This test wakes up (after the 1 sec), kills process_to_wait
step 7: monitor process seeks process_to_wait is dead, and kills the worker

google/datalab/contrib/mltoolbox/_shell_process.py
qimingj: The problem with shell=True is that it starts two processes (sh and python) and we cannot easily kill them.
I think "cd x&& python -m trainer.task --args" can work if we do "python -m path/to/trainer.task --args"? Or add the path to env?

tests/mltoolbox_magic/shell_process_tests.py
qimingj: Will add comments

google/datalab/contrib/mltoolbox/_shell_process.py
qimingj: We cannot do timeout here because the process we wait on is the kernel process. It should never time out. The use case is that user clicks the "reset session" button to kill the kernel process, and we should clean up all worker processes which this kernel started.

yebrahim: I see, makes sense.

google/datalab/contrib/mltoolbox/_shell_process.py
brandondutra: I just tested "python -m path/to/trainer.task", and I get the error "Import by filename is not supported." Looks like we'll have to set env vars.

qimingj: We can actually take a 'cwd' param and pass to subprocess.Popen. The following works:

p = subprocess.Popen(args, cwd='/content/dev/datalab/pydatalab/solutionbox/code_free_ml/mltoolbox/code_free_ml', env=os.environ)

qimingj: done.

tests/mltoolbox_magic/shell_process_tests.py
qimingj: Good idea. Added comments.

qimingj:
PTAL.

PR #406, opened 2017-06-11
Run and export test coverage numbers to coveralls

yebrahim:
The export is working fine, and the numbers show up on coveralls.io, but clicking a file to see its coverage details isn't working, giving the following error:

> **SOURCE NOT AVAILABLE**
> The file ".tox/coveralls/lib/python2.7/site-packages/google/datalab/ml/_summary.py" isn't available on github. Either it's been removed, or the repo root directory needs to be updated.
> 

It looks like it can't resolve the file path from inside the tox environment scope, but I'm not sure how to fix this. So for now, we can just run coverage locally to see the details.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,5 @@
+# .coveragerc to control coverage.py
+[run]
+
+[report]
+include = */site-packages/google/datalab/*
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 # By default, we want to run tests for Python 2.7, Python 3.5, and run our
 # flake8 checks.
-envlist = py27,py35,flake8
+envlist = py27,py35,flake8,coveralls
 # If an interpreter is missing locally, skip it.
 skip_missing_interpreters = true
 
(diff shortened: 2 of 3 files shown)


.coveragerc
```diff
@@ -0,0 +1,5 @@
+# .coveragerc to control coverage.py
+[run]
+
+[report]
+include = */site-packages/google/datalab/*
```
craigcitro: What if you include `*/google/datalab/*` and exclude `site-packages`? Does it figure out how to map the coverage info to the source files?

tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 # By default, we want to run tests for Python 2.7, Python 3.5, and run our
 # flake8 checks.
-envlist = py27,py35,flake8
+envlist = py27,py35,flake8,coveralls
```
yebrahim: Do you have a good pointer for this? It was painful setting up coveralls with tox due to the state of documentation, I'm not sure how to do this without a lot of trial and error.

craigcitro: Here's what I did for apitools:
  https://github.com/google/apitools/blob/master/tox.ini

Looking around, there seems to be a mix of models like that one, and people simply making coverage one of the tox envs, and then simply calling coveralls in the `after_success` section in travis.

.coveragerc
yebrahim: I don't think this works, `include` is used to filter the report with the provided pattern, it doesn't actually format the file paths.

craigcitro:
Looking around, it looks like this is something that lots of folks hit friction with:
  https://github.com/lemurheavy/coveralls-public/issues/263

It may be worth doing what you had originally and submitting, and then asking the coveralls folks.

yebrahim:
Yup, that's what I think we should do too. I fixed the script to ignore the coveralls command locally, but the coverage info is still collected, so you can do `coverage report` or `coverage html` locally to see the breakdown.

Now all I need is an approval. :)

yebrahim:
Ping for @ojarjur and @chmeyers, I need an owner approval too. :)

PR #407, opened 2017-06-12
Add more commands/utils tests

yebrahim:


google/datalab/utils/commands/_utils.py
```diff
@@ -461,7 +461,7 @@ def parse_control_options(controls, variable_defaults=None):
   div_id = _html.Html.next_id()
   if variable_defaults is None:
     variable_defaults = {}
-  for varname, control in list(controls.items()):
```
chmeyers: Does getting rid of the list() here make this non-Python3 compatible.  I feel like this might be one of the things that changed behaviors between 2 & 3, but I'm not sure...

yebrahim: Interesting, I didn't know about this. It looks like it's not required but I'll keep it for safety.

PR #408, opened 2017-06-13
MLToolbox: Add "%%ml analyze" magic command.

qimingj:


google/datalab/contrib/mltoolbox/commands/_ml.py
```diff
@@ -49,6 +57,78 @@ def ml(line, cell=None):
 Use "%ml <command> -h" for help on a specific command.
 """)
 
+  analyze_parser = parser.subcommand(
```
brandondutra: textwrap.dedent(""".... could be used to keep this block of text indented with the code, up to you.

qimingj: Will do in next change.

PR #409, opened 2017-06-13
allow one source column to be used in many feature columns

brandondutra:
this is done by adding a 'source_column' key to the transform dict in the features file. If this value is missing, the analyze script will automatically add it based on the key value. 

This also means tf-feature names do not have to have the same name as the schema/input file names. But a user would only see the tf-feature names in the exported model if they mark a feature as a key. 

A feature can be an input to a model and a key.

### DIFF
--- a/solutionbox/code_free_ml/mltoolbox/code_free_ml/analyze.py
+++ b/solutionbox/code_free_ml/mltoolbox/code_free_ml/analyze.py
@@ -420,26 +420,27 @@ def check_schema_transforms_match(schema, inverted_features):
     col_type = col_schema['type'].lower()
 
     # Check each transform and schema are compatible
-    for transform_name in inverted_features[col_name]:
-      if transform_name == constant.TARGET_TRANSFORM:
-        num_target_transforms += 1
-        continue
+    if col_name in inverted_features:
+      for transform_name in inverted_features[col_name]:
+        if transform_name == constant.TARGET_TRANSFORM:
+          num_target_transforms += 1
+          continue
 
-      elif col_type in constant.NUMERIC_SCHEMA:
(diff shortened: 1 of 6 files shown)


qimingj:
Update comments in analyze.py in the section that explains features file?

brandondutra:
added comments in analyze.py for the new feature file format

solutionbox/code_free_ml/mltoolbox/code_free_ml/analyze.py
```diff
@@ -158,6 +158,76 @@ def parse_arguments(argv):
   return args
 
 
+def make_analysis_plan(schema, features):
```
qimingj: Instead of introducing a concept "plan", I wonder if it is clearer if we structure the code such as:

1. Make a collections.DefaultDict(list), key being source column, value is the list of transforms.
2. Check for conflicting transforms in one place (check_schema_transforms_match(), or rename it).
3. During analysis, you only need to get the first transform in each source column since we know all transforms are compatible.

Thoughts?

brandondutra: The one thing I like about the plan concept is that we don't have to map each individual transform to an analysis step in two places (run_local_analysis and run_cloud_analysis), now it is done in one place (make_analysis_plan)

qimingj: Is the map as simple as the following?

if transform in CATEGORICAL_TRANSFORMS:
  ...
elif transform in NUMERIC_TRANSFORMS:
  ...

Then "plan" is not different from transforms group that you created in constants. When I looked at the code, it took me a while to figure out what "plan" is.

PR #412, opened 2017-06-14
More tests for utils._utils module

yebrahim:
Please let me know if there was a good reason we had the extra backslash in the windows path for gcloud.

### DIFF
--- a/tests/_util/util_tests.py
+++ b/tests/_util/util_tests.py
@@ -14,8 +14,13 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 import imp
 import unittest
+import pytz
+import mock
+import os
 
-from google.datalab.utils._utils import get_item
+import google.datalab.utils._utils as _utils
+from datetime import datetime
+import oauth2client.client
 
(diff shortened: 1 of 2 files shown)


ojarjur:
Regarding the backslash on Windows: That was added by https://github.com/googledatalab/datalab/pull/804, and there was no discussion at the time as to why it was added.

Unless we've explicitly tested this on Windows and confirmed that the backslash is unnecessary, I would just leave it.

tests/_util/util_tests.py
```diff
@@ -36,15 +41,144 @@ def _get_data():
...
+      mock_get_default_creds.side_effect = Exception
+      cred = _utils.get_credentials()
+
+      self.assertEquals(cred.access_token, 'test-access-token')
+
+    mock_path_exists.return_value = False
+    with self.assertRaises(Exception):
+      cred = _utils.get_credentials()
+
+    # If default creds are not define, and no file exists with credentials, throw
```
jimmc: "not defined"

PR #415, opened 2017-06-16
Add "%%ml train" sub command. Only local run is supported in this change.

qimingj:
Also modified mltoolbox's trainer/task.py to print better help string for model args.

### DIFF
--- a/google/datalab/contrib/mltoolbox/commands/_ml.py
+++ b/google/datalab/contrib/mltoolbox/commands/_ml.py
@@ -30,6 +30,7 @@ import pandas as pd
 import numpy as np
 import shutil
 import six
+import subprocess
 import tempfile
 
 import google.datalab
@@ -135,7 +136,7 @@ features: $features_def
       'transform', help="""Transform the data into tf.example which is more efficient
 in training. Example usage:
 
-%%ml transform --analysis_dir path/to/analysis --output_dir path/to/dir
+%%ml transform --analysis_output_dir path/to/analysis --output_dir path/to/dir
      --output_filename_prefix my_filename [--cloud] [--shuffle] [--batch_size 100]
 training_data:
   csv_file_pattern: path/to/csv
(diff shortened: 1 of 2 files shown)


google/datalab/contrib/mltoolbox/commands/_ml.py
```diff
@@ -330,6 +366,44 @@ def _transform(args, cell):
...
+
+  cmd_args = ['python', '-m', 'trainer.task',
+              '--job-dir', args['output_dir'],
+              '--analysis-output-dir', args['analysis_output_dir']]
+
+  def _process_train_eval_data(data, arg_name, cmd_args):
+    if isinstance(data, dict):
+      if 'csv_file_pattern' in data:
+        cmd_args.extend([arg_name, data['csv_file_pattern']])
+        cmd_args.append('--run-transforms')
```
brandondutra: I think this adds --run-transforms twice

google/datalab/contrib/mltoolbox/commands/_ml.py
```diff
@@ -330,6 +366,44 @@ def _transform(args, cell):
...
+      elif 'transformed_file_pattern' in data:
+        cmd_args.extend([arg_name, data['transformed_file_pattern']])
+      else:
+        raise ValueError('Invalid training_data dict. ' +
+                         'Requires either "csv_file_pattern" or "transformed_file_pattern".')
+    elif isinstance(data, google.datalab.ml.CsvDataSet):
+      cmd_args.extend([arg_name, data.input_files[0]])
```
brandondutra: This can be replaced with

for file_name in data.input_files:
  cmd_args.append(arg_name + '='+ file_name)

note 
1) more than one file can be given to the trainer
2)  We have to use '=' as the file name can have wildcard.

google/datalab/contrib/mltoolbox/commands/_ml.py
qimingj: fixed.

google/datalab/contrib/mltoolbox/commands/_ml.py
qimingj: yeah. Fixed and also updated transform and analyze.

qimingj:
PTAL.

PR #417, opened 2017-06-16
%%ml analyze can take a tarball 

brandondutra:
* add support for taking a tarball as the source
* will download the tarball if on gcs
* untar's it, and analyze uses the analyze.py file

todo

* support other file formats like .zip, .tar.bz, .tar.Z
* do this to transform and training too
* once we figure out the documentation story, we need to document the expected format of the tarball (analyze.py, transform.py, setup.py, trainer/task.py)

### DIFF
--- a/google/datalab/contrib/mltoolbox/commands/_ml.py
+++ b/google/datalab/contrib/mltoolbox/commands/_ml.py
@@ -38,6 +38,7 @@ import google.datalab.bigquery as bq
 import google.datalab.utils.commands
 import google.datalab.contrib.mltoolbox._local_predict as _local_predict
 import google.datalab.contrib.mltoolbox._shell_process as _shell_process
+import google.datalab.contrib.mltoolbox._archive as _archive
 
 
 MLTOOLBOX_CODE_PATH = '/datalab/lib/pydatalab/solutionbox/code_free_ml/mltoolbox/code_free_ml/'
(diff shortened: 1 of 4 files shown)


google/datalab/contrib/mltoolbox/commands/_ml.py
```diff
@@ -129,7 +130,8 @@ def ml(line, cell=None):
                               help='path of output directory.')
   analyze_parser.add_argument('--cloud', action='store_true', default=False,
                               help='whether to run analysis in cloud or local.')
-
+  analyze_parser.add_argument('--source_code', required=False,
```
qimingj: --package?

google/datalab/contrib/mltoolbox/commands/_ml.py
```diff
@@ -129,7 +130,8 @@ def ml(line, cell=None):
...
+                              help='A local or GCS tarball path to use as the source.')
```
qimingj: Also add that if missing, the default mltoolbox package is used?

google/datalab/contrib/mltoolbox/commands/_ml.py
brandondutra: done

google/datalab/contrib/mltoolbox/commands/_ml.py
brandondutra: done

brandondutra:
ptal

PR #419, opened 2017-06-19
Add cloud mode for "%%ml train" and "%%ml predict". Add "%%ml batch_predict".

qimingj:
Note:

Cloud batch prediction will come in a separate change.
_local_predict.py should be renamed to _predict.py since it contains also cloud prediction. It will happen in a future change.

### DIFF
--- a/google/datalab/contrib/mltoolbox/_local_predict.py
+++ b/google/datalab/contrib/mltoolbox/_local_predict.py
@@ -22,7 +22,6 @@ import collections
 import copy
 import csv
 from io import BytesIO
-import itertools
 import json
 import numpy as np
 import os
@@ -186,7 +185,7 @@ def get_prediction_results(model_dir_or_id, data, headers, img_cols=None,
 def _batch_csv_reader(csv_file, n):
   with file_io.FileIO(csv_file, 'r') as f:
     args = [f] * n
-    return itertools.izip_longest(*args)
+    return six.moves.zip_longest(*args)
 
 
 def _get_output_schema(session, output_alias_map):
@@ -206,7 +205,13 @@ def _get_output_schema(session, output_alias_map):
 
 def _format_results(output_format, output_schema, batched_results):
   # Convert a dict of list to a list of dict.
-  batched_results = [dict(zip(batched_results, t)) for t in zip(*batched_results.values())]
+  # Note that results from session.run may contain scaler value instead of lists
+  # if batch size is 1.
+  if isinstance(next(iter(batched_results.values())), list):
+    batched_results = [dict(zip(batched_results, t)) for t in zip(*batched_results.values())]
+  else:
+    batched_results = [batched_results]
+
   if output_format == 'csv':
     results = []
     for r in batched_results:
@@ -221,6 +226,8 @@ def _format_results(output_format, output_schema, batched_results):
             return int(obj)
         elif isinstance(obj, np.floating):
             return float(obj)
+        elif isinstance(obj, bytes):
+            return obj.decode('utf-8')
         elif isinstance(obj, np.ndarray):
             return obj.tolist()
         else:
@@ -233,7 +240,7 @@ def _format_results(output_format, output_schema, batched_results):
   return results
 
 
-def local_batch_predict(model_dir, csv_file_pattern, output_dir, output_format, batch_size=None):
+def local_batch_predict(model_dir, csv_file_pattern, output_dir, output_format, batch_size=100):
   """ Batch Predict with a specified model.
 
   It does batch prediction, saves results to output files and also creates an output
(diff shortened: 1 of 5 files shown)


google/datalab/contrib/mltoolbox/_local_predict.py
```diff
@@ -158,11 +165,99 @@ def get_prediction_results(model_dir, data, headers, img_cols=None, show_image=T
...
+    output_format: csv or json.
+  """
+  csv_files = file_io.get_matching_files(csv_file_pattern)
+  with tf.Graph().as_default(), tf.Session() as sess:
+    input_alias_map, output_alias_map = _tf_load_model(sess, model_dir)
+    csv_tensor_name = list(input_alias_map.values())[0]
+    output_schema = _get_output_schema(sess, output_alias_map)
+    with file_io.FileIO(output_file, 'w') as f:
+      for csv_file in csv_files:
+        prediction_source = _batch_csv_reader(csv_file, 100)
```
brandondutra: batch size needs to be expose. Maybe 100 examples will not fit in memory? Or add a todo for now

google/datalab/contrib/mltoolbox/_local_predict.py
```diff
@@ -158,11 +165,99 @@ def get_prediction_results(model_dir, data, headers, img_cols=None, show_image=T
...
+    csv_file_pattern: a pattern of csv files as batch prediction source.
+    output_file: the path of file to output.
+    output_format: csv or json.
+  """
+  csv_files = file_io.get_matching_files(csv_file_pattern)
+  with tf.Graph().as_default(), tf.Session() as sess:
+    input_alias_map, output_alias_map = _tf_load_model(sess, model_dir)
+    csv_tensor_name = list(input_alias_map.values())[0]
+    output_schema = _get_output_schema(sess, output_alias_map)
+    with file_io.FileIO(output_file, 'w') as f:
```
brandondutra: should the output be sharded if the input is? What if 100 1GB csv files are passed in? Maybe the number of shards could equal the number of input shards

google/datalab/contrib/mltoolbox/commands/_ml.py
```diff
@@ -369,41 +402,69 @@ def _transform(args, cell):
...
 
   if 'model_args' in cell_data:
     for k, v in six.iteritems(cell_data['model_args']):
-      cmd_args.extend(['--' + k, str(v)])
+      job_args.extend(['--' + k, str(v)])
 
-  _shell_process.run_and_monitor(cmd_args, os.getpid(), cwd=MLTOOLBOX_CODE_PATH)
+  if args['cloud']:
+    cloud_config = cell_data['cloud']
+    staging_tarball = os.path.join(args['output_dir'], 'staging', 'trainer.tar.gz')
```
brandondutra: check output_dir is on gcs?

google/datalab/contrib/mltoolbox/_local_predict.py
qimingj: Exposed it as an parameter of %%ml batch_predict.

google/datalab/contrib/mltoolbox/_local_predict.py
qimingj: Makes sense. Done.

google/datalab/contrib/mltoolbox/commands/_ml.py
qimingj: done.

qimingj:
PTAL

google/datalab/contrib/mltoolbox/_local_predict.py
```diff
@@ -158,11 +165,109 @@ def get_prediction_results(model_dir, data, headers, img_cols=None, show_image=T
...
+      schema.append({'name': name, 'type': 'FLOAT'})
+    else:
+      schema.append({'name': name, 'type': 'STRING'})
+
+  return schema
+
+
+def _format_results(output_format, output_schema, batched_results):
+  # Convert a dict of list to a list of dict.
+  batched_results = [dict(zip(batched_results, t)) for t in zip(*batched_results.values())]
```
brandondutra: does this work with batch_size = 1?

google/datalab/contrib/mltoolbox/_local_predict.py
```diff
@@ -158,11 +165,109 @@ def get_prediction_results(model_dir, data, headers, img_cols=None, show_image=T
...
+
+  Args:
+    model_dir: The model directory containing a SavedModel (usually saved_model.pb).
+    csv_file_pattern: a pattern of csv files as batch prediction source.
+    output_dir: the path of the output directory.
+    output_format: csv or json.
+    batch_size: if not specified, 100 is used. Larger batch_size improves performance but may
+        cause more memory usage.
+  """
+  if batch_size is None:
```
brandondutra: why not just put the 100 in the parameter list?

qimingj: Right. It was there to avoid an if else on the caller side. But maybe Datalab also support default parameter values (such as if --batch_size is missing, args['batch_size'] is 100 instead of None. Let me check if that works.

qimingj: yep. that works.

google/datalab/contrib/mltoolbox/_local_predict.py
qimingj: no. Nice catch! I didn't know TF returns individual values instead of list if batch size is 1. Fixed.

PR #420, opened 2017-06-19
tarball support for transform and train

brandondutra:


google/datalab/contrib/mltoolbox/_archive.py
```diff
@@ -36,6 +37,10 @@ def extract_archive(archive_path, dest):
 
   try:
     tmpfolder = None
+
+    if not (tf.gfile.Exists(archive_path) and not tf.gfile.IsDirectory(archive_path)):
```
qimingj: "if (not tf.gfile.Exists(archive_path)) or (tf.gfile.IsDirectory(archive_path)):" seems easier to understand.

google/datalab/contrib/mltoolbox/_archive.py
```diff
@@ -36,6 +37,10 @@ def extract_archive(archive_path, dest):
...
+      raise ValueError('archive path %s is not a file' % archive_path)
```
qimingj: "...does not exist or is not a file"

google/datalab/contrib/mltoolbox/commands/_ml.py
```diff
@@ -410,7 +414,19 @@ def _transform(args, cell):
   if '--project-id' not in cmd_args:
     cmd_args.extend(['--project-id', google.datalab.Context.default().project_id])
 
-  _shell_process.run_and_monitor(cmd_args, os.getpid(), cwd=MLTOOLBOX_CODE_PATH)
+  try:
+    tmpdir = None
```
qimingj: This works and I am okay with that. Is it possible to not delete the temp dir? For example, use a hash of the package URL as the dir name, so every time we delete the files in the dir first before downloading the package into the same dir. The benefits:

1. If user clicks "Reset" button, the temp dir will be cleaned up next time so at most we have one dir per package to keep on disk.

2. We don't need the try/finally blocks here.

google/datalab/contrib/mltoolbox/_archive.py
brandondutra: done

google/datalab/contrib/mltoolbox/_archive.py
brandondutra: done

google/datalab/contrib/mltoolbox/commands/_ml.py
brandondutra: I think we also need to hash the file, as the user could download the file and use a local path (run 1 notebook with 'file.tar.gz' from gs://bucket1/file.tar.gz, and another notebook at the same time  with 'file.tar.gz' form gs://bucket2/file.tar.gz).

This also doesn't address running the same package from two notebooks at the same time. When the 2nd notebook starts, it shouldn't overwrite files in the first one. Making a cache system is also a partial solution here. For example, what if the analyze.py script writes a file in dir_path(__file__).  The first notebook puts it in the cache, the 2nd notebook sees it already exists, and just runs it. So now two processes are writing to the same file.



Maybe we can make a cache.

All

PR #423, opened 2017-06-20
%%ml magic can take relative paths to files

brandondutra:
For example if --output_dir_from_analysis_step ./aout was used, the search would start from the source code folder, not the current folder because _shell_process.run_and_monitor() is ran in a different folder.

### DIFF
--- a/google/datalab/contrib/mltoolbox/commands/_ml.py
+++ b/google/datalab/contrib/mltoolbox/commands/_ml.py
@@ -306,9 +306,9 @@ def _abs_path(path):
   Args:
     path: string.
   """
-  if not path.startswith('gs://'):
-    return os.path.abspath(path)
-  return path
+  if path.startswith('gs://'):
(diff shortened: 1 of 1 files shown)


google/datalab/contrib/mltoolbox/commands/_ml.py
```diff
@@ -291,6 +291,26 @@ def ml(line, cell=None):
...
+  '../folder2/files*', then this function returns the string
+  '/content/datalab/folder2/files*'.
+
+  This function is needed if using _shell_process.run_and_monitor() as that
+  function runs a command in a different folder.
+
+  Args:
+    path: string.
+  """
+  if not path.startswith('gs://'):
```
qimingj: nit (save a not, and generally avoid not in if condition):

if path.startswith('gs://'):
  return path

return os.path.abspath(path)

PR #428, opened 2017-06-23
remove the printing lag

brandondutra:
some people think “Analyzing file /content/datalab/so/cleaned_train.csv” is printed out too late. and that I'm missing a sys.stdout.flush(). But I was already doing this!

Looking into this, I noticed the lag does not exist in the terminal version of the code, it only happens in the magic. The solution required two changes

1. in DL's subprocess, it should flush after printing
1. in analyze, output needs to have newlines, because p.stdout.readline() waits for a newline. This means we cannot do the "doing x..done" pattern. I now do "doing x...\n x is done" pattern.

I'm also printing when we are expanding the file pattern for local run as this can take a while if the files are on gcs.

### DIFF
--- a/solutionbox/code_free_ml/mltoolbox/code_free_ml/analyze.py
+++ b/solutionbox/code_free_ml/mltoolbox/code_free_ml/analyze.py
@@ -223,7 +223,7 @@ def run_cloud_analysis(output_dir, csv_file_pattern, bigquery_table, schema,
 
   numerical_vocab_stats = {}
   for col_name, transform_set in six.iteritems(inverted_features_target):
-    sys.stdout.write('Analyzing column %s...' % col_name)
+    sys.stdout.write('Analyzing column %s...\n' % col_name)
     sys.stdout.flush()
     # All transforms in transform_set require the same analysis. So look
     # at the first transform.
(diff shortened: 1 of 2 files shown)


solutionbox/code_free_ml/mltoolbox/code_free_ml/analyze.py
```diff
@@ -316,10 +316,14 @@ def run_local_analysis(output_dir, csv_file_pattern, schema, inverted_features):
   Raises:
     ValueError: on unknown transfrorms/schemas
   """
+  sys.stdout.write('Expanding any file patterns...\n')
```
qimingj: Do we need this printed? It seems expanding files should be quick.

brandondutra: not if the pattern is on GCS.

```
In [73]: %%time
    ...: file_io.get_matching_files('gs://cloud-ml-dev_bdt/*')
    ...: 
CPU times: user 1.36 s, sys: 128 ms, total: 1.49 s
Wall time: 12.2 s
Out[73]: 
['gs://cloud-ml-dev_bdt/analyze_out', ..., ]
```

brandondutra: ok, well, 12 secs is 'quick', but in terms of thinking printing 'analyze file' has lag, it's useful to show there is some startup costs.

qimingj: oh.. Makes sense.

PR #430, opened 2017-06-25
Schema validation when creating tables with repeating fields fails

jasadams:
As far as I know, when creating a table from a schema in BQ using datalab, if the schema has repeating fields, the schema validation fails.

After reading through the json schema documentation, I am under the impression that the table_schema_schema is incorrect and causing this validation error

This change makes "fields" type "array" with "#/definitions/field" items rather than type "#/definitions/field" in the json schema.

Be warned I have only tested the schema validation using the proposed schema directly within json-schema itself rather than in the datalab environment so this pull request requires further testing


jasadams:
I signed it!

yebrahim:
Thanks for the fix @jasadams. Can you also please add a simple test for this case, since it wasn't covered by our current tests. It should go in the [bigquery_tests module](https://github.com/googledatalab/pydatalab/blob/master/tests/kernel/bigquery_tests.py#L503).

PR #434, opened 2017-06-27
Add a feature in line/cell magic parsing, so that '--config1 value1' …

qimingj:
…can be specified in cell 'config1: value1'. This makes the notebook input pretty if the line args are too long.

I ran most of the sample notebooks to make sure it does not break existing notebooks, and also verified in these notebooks line param can be moved to cell too except for sql cell.

Note:

1. We do this only on the last subcommand. If your commands are like "%ml subcommand1 --config1 value1 subcommand2 --config2 value2", then it only tries getting "config2" from cell. But in reality we don't have those cases that there are arguments in multiple subcommands.

2. Only param beginning with '--' is supported.


qimingj:
The test coverage details are confusing (https://coveralls.io/builds/12143674). The change does not touch the first 3 files listed there, so I don't know why the test coverage changes for these files.

qimingj:
NVM. It was because I removed 2 tests accidentally from the list. Will add them back.

google/datalab/utils/commands/_utils.py
```diff
@@ -336,6 +338,62 @@ def parse_config(config, env, as_dict=True):
...
+  elif stripped[0] == '{':
+    config = json.loads(content)
+  else:
+    config = yaml.load(content)
+
+  if not isinstance(config, dict):
+    raise ValueError('Invalid config.')
+
+  config_items = {k: v for k, v in six.iteritems(config)
+                  if k in keys and (isinstance(v, six.string_types) or isinstance(v, bool))}
```
brandondutra: why not just 'k in keys'? 

Why can't v be an int?

google/datalab/utils/commands/_utils.py
```diff
@@ -336,6 +338,62 @@ def parse_config(config, env, as_dict=True):
...
+  and 'keys' are: [config_item1, config_item3]
+
+  The results will be a tuple of
+  1. The parsed config items (dict): {config_item1: value1, config_item3: value3}
+  2. The remaining content (string): config_item2: value2
+
+  Args:
+    content: the input content. A string. It has to be a yaml or JSON string.
+    env: user supplied dictionary for replacing vars (for example, $myvalue).
+    keys: a list of keys to retrieve from content.
```
brandondutra: add:

keys must be a top-level key. So if content is  
config_item4:
   config_item5: True

only config_item4 could be a key

google/datalab/utils/commands/_utils.py
```diff
@@ -336,6 +338,62 @@ def parse_config(config, env, as_dict=True):
...
+  config_items = {k: v for k, v in six.iteritems(config)
+                  if k in keys and (isinstance(v, six.string_types) or isinstance(v, bool))}
+  config_remaining = {k: v for k, v in six.iteritems(config) if k not in config_items}
+  replace_vars(config_items, env)
+
+  if not config_remaining:
+    return config_items, None
+
+  if stripped[0] == '{':
+    content_items_removed = json.dumps(config_remaining)
```
brandondutra: if using default_flow_style below, then json should use indent

tests/_util/commands_tests.py
```diff
@@ -0,0 +1,76 @@
...
+
+    args, cell = parser.parse('subcommand1 --string1 value1', 'string3: value3')
+    self.assertEqual(vars(args), {'string1': 'value1', 'flag1': False})
+    self.assertEqual(cell.strip(), 'string3: value3')
+
+    args, cell = parser.parse('subcommand1', 'string1: value1\nflag1: true')
+    self.assertEqual(vars(args), {'string1': 'value1', 'flag1': True})
+    self.assertIsNone(cell)
+
+    parser = CommandParser(
```
brandondutra: the subcommand test can be put into a new test function. So test errors are easier to debug.

google/datalab/utils/commands/_commands.py
```diff
@@ -63,17 +76,88 @@ def create_args(line, namespace):
...
+        subcommand_name_to_args[name] = args_to_parse
+      self._subparser_args_to_parse_from_cell = subcommand_name_to_args
+    return self._subparser_args_to_parse_from_cell
+
+  def parse(self, line, cell, namespace=None):
+    """Parses a line and cell into a dictionary of arguments, expanding variables from a namespace.
+
+    For each line parameters beginning with --, it also checks the cell content and see if it exists
+    there. For example, if "--config1" is a line parameter, it checks to see if cell dict contains
+    "config1" item, and if so, use the cell value. The "config1" item will also be removed from
```
yebrahim: I think this might cause some misunderstanding. How about we throw an error instead, saying that the parameter is specified in two different ways? Otherwise I'm not sure which should take the precedence.

google/datalab/utils/commands/_utils.py
```diff
@@ -336,6 +338,62 @@ def parse_config(config, env, as_dict=True):
   return config
 
 
+def parse_config_for_selected_keys(content, env, keys):
```
yebrahim: Tests for this method?

qimingj:
One reason to allow line params in cell is that existing magics get it for free. For example, in
https://github.com/googledatalab/notebooks/blob/master/tutorials/BigQuery/BigQuery%20Commands.ipynb, we have:

%bq sample --table cloud-datalab-samples.httplogs.logs_20140615 --count 10 --fields timestamp,latency --method hashed --key-field latency

And now it can be:

%bq sample
table: cloud-datalab-samples.httplogs.logs_20140615
count: 10
fields: timestamp,latency
method: hashed
key-field: latency

If we explicitly force the separation of line and cell params, we will have to touch all the magics to move line params to cell, and it will be a breaking change.

google/datalab/utils/commands/_commands.py
```diff
@@ -63,17 +76,88 @@ def create_args(line, namespace):
...
     return args
 
-  def parse(self, line, namespace=None):
-    """Parses a line into a dictionary of arguments, expanding meta-variables from a namespace. """
+  def _get_args_to_parse_from_cell(self):
+
+    def _get_subparsers(p):
+      """Recursively get subparsers."""
+
+      subparsers = [
```
qimingj: done.

google/datalab/utils/commands/_commands.py
```diff
@@ -63,17 +76,88 @@ def create_args(line, namespace):
...
+        name = subparser.prog.split()[-1]
+        args_to_parse = subparser._args_to_parse_from_cell
+        subcommand_name_to_args[name] = args_to_parse
+      self._subparser_args_to_parse_from_cell = subcommand_name_to_args
+    return self._subparser_args_to_parse_from_cell
+
+  def parse(self, line, cell, namespace=None):
+    """Parses a line and cell into a dictionary of arguments, expanding variables from a namespace.
+
+    For each line parameters beginning with --, it also checks the cell content and see if it exists
```
qimingj: Now we don't allow same arg provided in both cell and line.

google/datalab/utils/commands/_commands.py
```diff
@@ -63,17 +76,88 @@ def create_args(line, namespace):
...
+      ipy = IPython.get_ipython()
+      namespace = ipy.user_ns
     try:
-      if namespace is None:
-        ipy = IPython.get_ipython()
-        namespace = ipy.user_ns
       args = CommandParser.create_args(line, namespace)
-      return self.parse_args(args)
+      subparser_name_to_args = self._get_args_to_parse_from_cell()
+      last_subcmd = next((x for x in reversed(args) if x in subparser_name_to_args), None)
```
qimingj: Because we don't want to get in the business of line args insertion. For simplicity, we always append at the end of the line.

None of our usage involves multiple subcommands in the same input.

google/datalab/utils/commands/_commands.py
qimingj: I agree. Done.

google/datalab/utils/commands/_utils.py
qimingj: Will do that in a separate change. For now it is partially covered in CommandParser tests.

tests/_util/commands_tests.py
qimingj: done.

google/datalab/utils/commands/_utils.py
qimingj: done.

google/datalab/utils/commands/_utils.py
qimingj: This piece of code is removed but yes, good catch.

google/datalab/utils/commands/_utils.py
qimingj: done.

qimingj:
PTAL. Thanks!

google/datalab/utils/commands/_commands.py
```diff
@@ -63,17 +83,156 @@ def create_args(line, namespace):
...
+          if arg_value is None:
+            continue
+
+          if '--' + arg_name in args:
+            raise ValueError('config item "%s" is specified in both cell and line.' % arg_name)
+          if isinstance(arg_value, bool):
+            if arg_value:
+              line += ' --%s' % arg_name
+          else:
+            line += ' --%s %s' % (arg_name, str(cell_config[arg_name]))
```
brandondutra: does this work with --cloud?

%%ml transform
cloud:
  num_workers: 10
  

This means that --cloud cannot be a line arg.

google/datalab/utils/commands/_commands.py
```diff
@@ -22,15 +22,22 @@
...
   """An argument parser to parse commands in line/cell magic declarations. """
 
   def __init__(self, *args, **kwargs):
     """Initializes an instance of a CommandParser. """
+
     super(CommandParser, self).__init__(*args, **kwargs)
+    # Set _parser_class, so that subparsers added will also be of this type.
+    self._parser_class = CommandParser
     self._subcommands = None
+    self._cell_args = {}
```
qimingj: done.

google/datalab/utils/commands/_commands.py
```diff
@@ -63,17 +83,156 @@ def create_args(line, namespace):
...
+
+    subparsers = self._get_subparsers()
+    for subparser in subparsers:
+      if subparser_prog == subparser.prog:
+        # Found the subparser.
+        args_to_parse = []
+        for action in subparser._actions:
+          if (isinstance(action, argparse._StoreAction) or
+             isinstance(action, argparse._StoreTrueAction) or
+             isinstance(action, argparse._StoreFalseAction)):
```
qimingj: Nice idea. Done.

google/datalab/utils/commands/_commands.py
```diff
@@ -63,17 +83,156 @@ def create_args(line, namespace):
...
+    # Find which subcommand in the line by comparing line with subcommand progs.
+    # For example, assuming there are 3 subcommands with their progs
+    #   %bq tables
+    #   %bq tables list
+    #   %bq datasets
+    # and the line is "tables list --dataset proj.myds"
+    # it will find the second one --- "tables list" because it matches the prog and
+    # it is the longest.
+    args = CommandParser.create_args(line, namespace)
+    sub_parsers_progs = [x.prog for x in self._get_subparsers()]
```
qimingj: added comments.

google/datalab/utils/commands/_commands.py
```diff
@@ -63,17 +83,156 @@ def create_args(line, namespace):
...
+    matched_progs = []
+    for prog in sub_parsers_progs:
+      # Remove the leading magic such as "%bq".
+      match = prog.split()[1:]
+      if next((match for i in range(len(args)) if args[i:i + len(match)] == match), []):
```
qimingj: done

google/datalab/utils/commands/_commands.py
qimingj: No. It doesn't. I will make another change so "cloud" in cell becomes "cloud_config".

google/datalab/utils/commands/_utils.py
```diff
@@ -249,13 +249,15 @@ def get_data(source, fields='*', env=None, first_row=0, count=-1, schema=None):
...
-      return args.func(vars(args), cell)
-    except Exception as e:
-      sys.stderr.write(str(e))
-      sys.stderr.write('\n')
+  try:
+    args, cell = parser.parse(line, cell, namespace)
+    if args:
+      return args['func'](args, cell)
+  except Exception as e:
+    # e.args[0] is None if --help is provided in line.
```
qimingj: Agree. To make things simple both sides will agree on a string 'exit_0'.

tests/_util/commands_tests.py
```diff
@@ -0,0 +1,153 @@
...
+    self.assertEqual(args, {'string1': 'value1', 'string2': 'value2', 'string3': None,
+                            'command': 'subcommand2', 'flag1': True})
+    self.assertIsNone(cell)
+
+    args, cell = parser.parse('subcommand1 subcommand2',
+                              'string1: value1\nstring2: value2\nstring3: value3\n' +
+                              'string4: value4\nflag1: false')
+    self.assertEqual(args, {'string1': 'value1', 'string2': 'value2', 'string3': 'value3',
+                            'command': 'subcommand2', 'flag1': False})
+    self.assertEqual(cell.strip(), 'string3: value3\nstring4: value4')
```
qimingj: done.

tests/_util/commands_tests.py
```diff
@@ -0,0 +1,153 @@
...
+
+    with self.assertRaises(ValueError):
+      subcommand2.add_argument('--duparg', help='help.')
+      subcommand2.add_cell_argument('duparg', help='help.')
+
+    with self.assertRaises(ValueError):
+      parser.parse('subcommand1 subcommand2 -s value1 --duparg v1', 'duparg: v2')
+
+    with self.assertRaises(Exception):
+      parser.parse('subcommand1 subcommand2 -s value1 --string3 value3', 'a: b')
```
qimingj: done.

tests/_util/commands_tests.py
```diff
@@ -0,0 +1,153 @@
...
+    self.assertEqual(cell.strip(), 'a: b\nstring2: $var2')
+
+    cell = """
+dict1:
+  k1: $var1
+  k2: $var3
+"""
+    args, cell = parser.parse('subcommand1', cell, namespace)
+    self.assertEqual(args['dict1'], {'k1': 'value1', 'k2': [1, 2]})
+
```
qimingj: The first test(test_subcommand_line) covers it.

qimingj:
Thanks for the catches! PTAL.

PR #435, opened 2017-06-27
fix evaluation issue

brandondutra:
This fixed the zero accuracy/wrong accuracy on my test data. However, b/62844938 says this is not the complete story. In any event, allow_smaller_final_batch=True is the correct thing to do for eval. 

For training, it is a little more complicated. If false and the batch size is large and num_epochs is used, then not all the data is used. If true and using num_epochs, the last batch could have very few elements (like 1), but the learning rate depends on batch size, so the model would overfit on the last example. I went with true for training.


qimingj:
Makes sense. In reality there may not be many training instances. In an extreme case, if instances = 180 and batch size is 100, we don't want to lose the 80 instances.

PR #439, opened 2017-06-28
add sample test

brandondutra:
csv dataset sample bug was fixed in #397. A user hit this bug today, and I noticed we don't have a test for it. So adding a test now.

### DIFF
--- a/tests/ml/dataset_tests.py
+++ b/tests/ml/dataset_tests.py
@@ -14,14 +14,20 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 import unittest
 
+import csv
+import os
+import random
+import shutil
+import tempfile
+
 import google.datalab
 from google.datalab.ml import CsvDataSet
 
 
-class TestCases(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


tests/ml/dataset_tests.py
```diff
@@ -35,6 +41,48 @@ def test_create_dataset(self):
...
 
+  def test_sample(self):
+    tmp_dir = tempfile.mkdtemp()
+    try:
+      json_schema = TestCsvDataSet._create_json_schema()
+      TestCsvDataSet._create_csv_files(tmp_dir, 'data', 3)
+      ds = CsvDataSet(file_pattern=os.path.join(tmp_dir, 'data*'), schema=json_schema)
+
+      df = ds.sample(5)
+      self.assertEqual(5, len(df))
```
qimingj: Also assert the following?

- the 5 samples belong to csv file,
- they are different (so we may need to generate a unique key for each row)
- Returned type is dataframe

brandondutra: done!

PR #441, opened 2017-06-28
%%ml predict works with one prediction example + print trace

brandondutra:
our model calls tf.squeeze on the output tensors, which removes 1 dim values. So something like [[a], [b]] becomes [a, b]. But this also means [[a]] becomes a, which looks nice but means we have to deal with batch size 1 as a special case.

This PR also prints the stack trace when an exception is raised in any magic command in DL.


brandondutra:
per review comments, reverted the stack trace change as in the future we'll have an --verbose or --debug flag on all magics to print stack traces.

google/datalab/contrib/mltoolbox/_local_predict.py
```diff
@@ -53,14 +53,35 @@ def _tf_load_model(sess, model_dir):
 
 
 def _tf_predict(model_dir, input_csvlines):
-  """Prediction with a tf savedmodel."""
+  """Prediction with a tf savedmodel.
+
+  Args:
+    model_dir: folder that contains a saved model
```
qimingj: nit: folder -> directory

google/datalab/contrib/mltoolbox/_local_predict.py
```diff
@@ -53,14 +53,35 @@ def _tf_load_model(sess, model_dir):
...
     input_alias_map, output_alias_map = _tf_load_model(sess, model_dir)
     csv_tensor_name = list(input_alias_map.values())[0]
     results = sess.run(fetches=output_alias_map,
                        feed_dict={csv_tensor_name: input_csvlines})
 
+  # convert any scalar values to a list. This may happen when there is one
+  # example in input_csvlines and the model uses tf.squeeze on the output
+  # tensor.
+  if len(input_csvlines) == 1:
+    updated_results = {}
```
qimingj: nit: can we reuse "results" here so we can save line 77 and line 83?

google/datalab/contrib/mltoolbox/_local_predict.py
brandondutra: done

PR #446, opened 2017-06-30
Add tests to code_free_ml package

brandondutra:
I added the code_free_ml tests to travis. 

Some notes

* the pydatalab repo cannot use GCP resources currently. So some tests are skipped in travis. I'll connect with Omar later and see if this can be removed.
* the added tests only run in python 2 because of dataflow
* also fixed some flaky things with the code_free_ml tests

Tests have to be run via a subprocess command as code_free_ml cannot be installed and ran like a normal package. As an example, if we installed the code_free_ml package like how we do for the other mltoolbox pacakges, we cannot run 'mltoolbox.code_free_ml.analyze.main()' as that file imports from trainer. This works when analyze.py is ran as a script, but fails because we don't  install it's trainer module. This is fine as analyze.py should only be ran as a script and not as a module.


yebrahim:
Why is coverage going down by so much with this change?

brandondutra:
Don't worry about it, it is not ready for review. I know why.

brandondutra:
ready for review, coverage is back to normal. If this PR adds test, why did it not increase? Because the coverage report does not check the mltoolbox packages.

solutionbox/code_free_ml/test_mltoolbox/test_analyze.py
```diff
@@ -226,8 +235,11 @@ def test_text(self):
       vocab = pd.read_csv(six.StringIO(vocab_str),
                           header=None,
                           names=['col1', 'count'])
-      self.assertEqual(vocab['col1'].tolist(),
-                       ['quick', 'brown', 'the', 'fox', 'chicken'])
+
+      # vocabs are sorted by count only
+      col1_vocab = vocab['col1'].tolist()
+      self.assertItemsEqual(col1_vocab[:2], ['brown', 'quick'])
```
yebrahim: I'm curious what this adds over just comparing the entire list once?

brandondutra:
ptal

tests/mltoolbox_code_free_ml/all_tests.py
```diff
@@ -0,0 +1,45 @@
...
+    self._root_path = os.path.abspath(
+        os.path.join(
+            os.path.dirname(__file__),
+            '..', '..', 'solutionbox', 'code_free_ml', 'test_mltoolbox'))
+
+  @unittest.skipIf(not six.PY2, 'Python 2 is required')
+  def test_local(self):
+    """Run some of the code_free_ml tests.
+
+    Tests that use GCS services like GCS or BigQuery are not ran.
```
yebrahim: typo: "run".

tests/mltoolbox_code_free_ml/all_tests.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    cmd = 'bash %s' % os.path.join(self._root_path, 'run_all.sh')
+    subprocess.check_call(cmd, cwd=self._root_path, shell=True)
```
yebrahim: Is there no way we can run these tests directly by calling them from python? I just would like to reduce the friction of adding coverage for all this later.

brandondutra: The problem is that in analyze.py, it needs to run code in trainer/feature_transforms.py

When analyze is called like a script, "from trainer import feature_transforms " works. But if you install the code_free_ml package and use 'from mltoolbox.code_free_ml import analyze', you get an import error saying No module named trainer.

PR #450, opened 2017-07-03
Fix MLToolbox transform service run. Also add a check to make sure if cloud_config is specified, --cloud must be there.

qimingj:


brandondutra:
Bradley is OOO. When tests are finished, I'll merge.

PR #451, opened 2017-07-05
Fix extract parse bug

yebrahim:
Also a minor bug with the environment variable getter.

Fixes https://github.com/googledatalab/datalab/issues/1486

### DIFF
--- a/google/datalab/bigquery/commands/_bigquery.py
+++ b/google/datalab/bigquery/commands/_bigquery.py
@@ -787,7 +787,8 @@ def _extract_cell(args, cell_body):
                          csv_delimiter=args['delimiter'], csv_header=args['header'],
                          compress=args['compress'])
   elif args['query'] or args['view']:
-    source = google.datalab.utils.commands.get_notebook_item(args['view'])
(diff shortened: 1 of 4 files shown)


google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -787,7 +787,8 @@ def _extract_cell(args, cell_body):
                          csv_delimiter=args['delimiter'], csv_header=args['header'],
                          compress=args['compress'])
   elif args['query'] or args['view']:
-    source = google.datalab.utils.commands.get_notebook_item(args['view'])
+    source_name = args['view'] if args['view'] else args['query']
```
ojarjur: This feels redundant to me.

What about using `source_name = args['view'] or args['query']` instead?

parthea:
Is it possible to write a unit test that fails without the fix?

google/datalab/bigquery/commands/_bigquery.py
yebrahim: done.

yebrahim:
Fixed the unit tests to cover this case.

PR #453, opened 2017-07-07
make pydatalab use tf 1.2.0

brandondutra:
many  changes:

* tf estimator's feature_engineering_fn broke somehow. But I learned support for this is being removed anyways. So the work that function did (apply a hidden layer to image embedding columns) moved to the training/serving input functions. This required a new serving function for the transform step as the transform step should not have a hidden layer.

* tf 1.2.0 does a better job of str vs bytes for python3. I ended up removing the old mltoolbox code from the python3 tests. Before, non-dataflow steps were being tested in python 3, but this is silly as mltoolbox cannot be used without dataflow and hence must be python 2.

* protobuf is still used in setup.py files. This is because tf has to be installed on Dataflow workers. 

* remove string_to_index_table_from_tensor warning

* updated make_export_strategy to sync with export_savedmodel. I explored Eli's trick in https://stackoverflow.com/questions/44381879/training-and-predicting-with-instance-keys However, because of the target/no target exported models, my export function would need to change the model. I need to connect with Eli more, but I don't see how we can just call the estimator's export_savedmodel. Also updated make_export_strategy in _structure_data mltoolbox package.

* bump old mltoolbox package version numbers

### DIFF
--- a/google/datalab/contrib/mltoolbox/_local_predict.py
+++ b/google/datalab/contrib/mltoolbox/_local_predict.py
@@ -88,7 +88,7 @@ def _download_images(data, img_cols):
   for d in data:
     for img_col in img_cols:
       if d.get(img_col, None):
-        with file_io.FileIO(d[img_col], 'r') as fi:
+        with file_io.FileIO(d[img_col], 'rb') as fi:
           im = Image.open(fi)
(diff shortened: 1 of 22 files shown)


brandondutra:
ptal

PR #459, opened 2017-07-15
Fix an issue that when variable replacement fails in cell input, it does not give proper error message.

qimingj:
The fix moves replace_vars() out of the try block, so if replace_vars() fails the error will show up.

The change in _bigquery.py is needed because otherwise an empty (but not None) namespace is passed to parser, and if there are any variable ref in cell input, it will fail to resolve. Also, handle_magic_line() takes care of namespace if it is None, so no need to explicitly pass in a namespace.

### DIFF
--- a/google/datalab/bigquery/commands/_bigquery.py
+++ b/google/datalab/bigquery/commands/_bigquery.py
@@ -929,13 +929,7 @@ def bq(line, cell=None):
   Use %bq --help for a list of commands, or %bq <command> --help for help
   on a specific command.
   """
-  namespace = {}
-  if line.find('$') >= 0:
-    # We likely have variables to expand; get the appropriate context.
-    namespace = google.datalab.utils.commands.notebook_environment()
-
-  return google.datalab.utils.commands.handle_magic_line(line, cell, _bigquery_parser,
-                                                         namespace=namespace)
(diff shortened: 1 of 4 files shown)


google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -929,13 +929,7 @@ def bq(line, cell=None):
...
   on a specific command.
   """
-  namespace = {}
-  if line.find('$') >= 0:
-    # We likely have variables to expand; get the appropriate context.
-    namespace = google.datalab.utils.commands.notebook_environment()
-
-  return google.datalab.utils.commands.handle_magic_line(line, cell, _bigquery_parser,
-                                                         namespace=namespace)
+  return google.datalab.utils.commands.handle_magic_line(line, cell, _bigquery_parser)
```
yebrahim: Where is the namespace getting passed if not here?

qimingj: If namespace is not passed, it uses IPython's user namespace. See https://github.com/googledatalab/pydatalab/blob/master/google/datalab/utils/commands/_commands.py#L174.

yebrahim: LGTM then.

PR #460, opened 2017-07-17
Add test for nested schema

yebrahim:
Following up from https://github.com/googledatalab/pydatalab/pull/430.

### DIFF
--- a/tests/kernel/bigquery_tests.py
+++ b/tests/kernel/bigquery_tests.py
@@ -48,7 +48,7 @@ class TestCases(unittest.TestCase):
 
   def test_table_schema(self):
     import jsonschema
-    schemas = [
+    good_schemas = [
       {
         'schema': [
           {'name': 'col1', 'type': 'int64', 'mode': 'NULLABLE', 'description': 'description1'},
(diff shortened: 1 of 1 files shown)


tests/kernel/bigquery_tests.py
```diff
@@ -46,6 +46,35 @@ def noop_decorator(func):
...
+                {'name': 'nestedField2', 'type': 'STRING'}
+              ]}
+           ]}
+        ]
+      }
+    ]
+
+    for s in schemas:
+      record = google.datalab.utils.commands.parse_config(json.dumps(s), {})
+      jsonschema.validate(record, google.datalab.bigquery.commands._bigquery.table_schema_schema)
```
jimmc: I don't understand what this test is supposed to be testing. It seems like this test should include some good schemas and some bad schemas, but I can't tell from the code whether that is the intent.

yebrahim: This makes sure some good scenarios are valid. Schemas with nested fields were broken until #430, so this is a regression test.

yebrahim: Added some bad schemas for basic scenarios.

jimmc: Thanks, that helps. It is counter-intuitive that a schema that has two 'col1' columns with different types is valid.

yebrahim: Yes this is just a structural integrity test. Bigquery will reject the schema at the service, but we don't want to replicate all their tests on the client.

PR #464, opened 2017-07-19
Add facets visualization components. 

qimingj:
See https://github.com/PAIR-code/facets for facets. This enables feature exploration in ML story.

The python files under google/datalab/utils/facets are copied from facets source repo. This is a temp workaround until facets is available in pypi.


brandondutra:
their github page says "The Facets visualizations currently work only in Chrome".  Does this mean I cannot use firefox with datalab now?

brandondutra:
how about google/datalab/utils/_facets so users don't import it

qimingj:
Hmm. I didn't noticed that. Maybe it was not well tested on other browsers. Given that majority of datalab users use chrome, I am not concerned about it.

qimingj:
how about google/datalab/utils/_facets so users don't import it:

Good idea. Will do.

qimingj:
wait... How do I import it then?

brandondutra:
I'm worried about the Chrome thing. Can you test it works in firefox/safari when using a VM.

brandondutra:
google/datalab/ml/_fasets.py has the classes a user imports, right? So maybe this file should be

google/datalab/ml/fasets.py

or maybe even

google/datalab/facets/_fasets.py

brandondutra:
sorry, if this is a temp workaround, then you don't what users to import it at all [if we assume this will be in pypi one day]. So is the plan to wrap this in a magic?

qimingj:
I see the confusion. My bad. Files under google/datalab/utils/facets are helper libraries and will be removed once facets is available in pypi. They are importable because google/datalab/ml/facets need them. What users want to import is FacetsOverview and FacetsDiveview under google.datalab.ml. I forgot to make the two classes importable which leads to confusion (__init__.py updated).

It doesn't work on firefox so far. Maybe we should add comments in the class docstring? It doesn't error out anyway, just no output.

+ 85% of datalab users are chrome users.

qimingj:
I opened issue https://github.com/PAIR-code/facets/issues/26 to track firefox support.

qimingj:
Fixed some py3 issues. PTAL.

PR #466, opened 2017-07-22
top_n=0 means all the classes

brandondutra:


brandondutra:
ptal, a push removed your github comment :(

PR #468, opened 2017-07-25
versions support runtime version

brandondutra:


google/datalab/ml/_cloud_models.py
```diff
@@ -168,12 +168,13 @@ def get_version_details(self, version_name):
     name = ('%s/versions/%s' % (self._full_model_name, version_name))
     return self._api.projects().models().versions().get(name=name).execute()
 
-  def deploy(self, version_name, path):
+  def deploy(self, version_name, path, runtime_version='1.0'):
```
qimingj: Should the default be "not set"? I think if not set, CloudML will choose a version.

brandondutra: it was currently set to use "1.0" always, so I didn't change the default behavior. Do you want this changed to "not set", which might break some people but could provide a better experience. Do you agree with this tradeoff?

qimingj: This might hurt future users. Imagine when tf 2.0 is out is becoming mainstream, and many users don't care to set the runtime version. It may cause more problems. We cannot change the default value to '2.0' again.

Setting it to 1.0 was a mistake. I think we can still recover when 1.0 and 1.2 are still close.

brandondutra: agree, will change

brandondutra: Fixed in #469

PR #481, opened 2017-08-18
Initial support for Datalab widgets in sandboxed outputs.

blois:
This adds support for running Datalab widgets in sandboxed browser environments where the global window environment has not been populated with any of the regular Datalab state.

An open question is when this code should be invoked- it can be something which the user needs to explicitly opt into by invoking from their notebook code, or it could be invoked automatically since it's basically a pre-req for some of the visualizations.


yebrahim:
Looks like you have some simple lint errors, PTAL @blois.

google/datalab/utils/_utils.py
```diff
@@ -263,3 +263,50 @@ def python_portable_string(string, encoding='utf-8'):
     return string.decode(encoding)
 
   raise ValueError('Unsupported type %s' % str(type(string)))
+
+
+_sandboxed_output_hook = None
+def initialize_sandboxed_outputs():
```
yebrahim: Top level functions need two blank lines before them. Also on line 301.

blois:
Updated!

blois:
Updated CLA for @peteblois.

yebrahim:
Closing this in favor of https://github.com/googledatalab/pydatalab/pull/517.

PR #482, opened 2017-08-19
Bump version up to 1.1.0

### LINKED ISSUE #422: Update the homepage link on PyPi
The homepage link on https://pypi.python.org/pypi/datalab/1.0.1 is currently pointing to https://github.com/googledatalab/datalab but I think it should be https://github.com/googledatalab/pydatalab

yebrahim:
Bumping version, and updating the release script, since instructions have changed with the introduction of the new pypi.org package index.

Also fixes #422.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@
 
 from setuptools import setup
 
-version = '1.0.1'
+version = '1.1.0'
 
 setup(
   name='datalab',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -14,7 +14,7 @@
 
 from setuptools import setup
 
-version = '1.0.1'
+version = '1.1.0'
```
qimingj: Our current datalab version is 1.2.*. Maybe pydatalab should be 1.2.0 too?

yebrahim: I don’t see a reason why we should tie these two together. We should be able to iterate on pydatalab individually.

qimingj: My reason is consistency. pydatalab is part of datalab so it is easier to use same version to avoid confusions (users don't care about pydatalab/datalab releases). I don't see a case that we have enough reasons to release a new pydatalab, but not datalab image. Since we are now releasing them in the same cadence, it is a good opportunity to align the versions.

qimingj: Let's keep it 1.1 and discuss later how to manage pydatalab releasing.

PR #480, opened 2017-08-18
Introducing Datalab Notebook Productionization with %pipeline

rajivpb:
Usage:
%pipeline create --name foo_pipeline
<yaml config for pipeline>

This also introduces the 'create' subcommand. The implementation of this generates the python airflow spec and stores it in memory. A few of the python fields are hard-coded for now and these will be parameterized eventually.

PR #484, opened 2017-08-23
Introducing Datalab Notebook Productionization with %pipeline (#480)

rajivpb:
* Airflow python spec generation.

* Airflow python spec generation

* Deleting extraneous files from tests

* Removing TODO to pick either Yaml or Json. We'll allow both. No harm in that.

* Adding ability to specify 'query' instead of 'bql' for BigQuery.

* airflow_tests were being run twice in error. Fixing.

* Auto merged files after git rebase

* Resolving test conflicts.

* Fixing up test related changes

* Making new top-level schedule section in YAML spec; AirflowDag->AirflowPipeline rename

* dag_id -> pipeline_id

* Introducing BigQueryTableDeleteOperator, BigQueryToBigQueryOperator, and BigQueryToCloudStorageOperator

* _airflow_dag.py -> _airflow_pipeline.py

* Introducing %pipeline magic with support for "create" command. Also, some refactoring: AirflowPipeline -> Pipeline. Pipeline -> Deploy.

* Deleting airflow_job.py

* Importing the %%pipeline magic to kernels so that it's callable from datalab.

* Plugging into datalab object model to parse "pipeline create" magic.

* Stub implementation for pipeline load functionality

* BQ queries can now be referenced by name in the yaml spec!

* Some refactoring / clean-up.

* airflow -> pipeline

* Using datalab's parse_config for cell_body parsing (and throwing out _coders.py)

* Airflow spec now does not specify default params if the values are defaults.

* Removing "deploy" subcommand. This will be folded into create and there will be a single subcommand under the %pipeline magic.

* Some cleanup ahead of PR.

* Minor cleanup

* Adding doc strings, and fixing tests because python 3 renamed dict.iteritems -> dict.items

* More doc-strings

* Sorting dictionaries after iteration so that the resulting order is stable.

* Fixing lint issues

* Fixing flake8 errors (still a few to go).

* Fixing/suppressing remaining flake8 errors

* Fixing year (2015->2017) in all __init__.py files

* Responding to code-review suggestions


qimingj:
Since it is going to master now, I would suggest not enabling %%pipeline by default and have it live under contrib namespace. See %%ml (https://github.com/googledatalab/pydatalab/tree/master/google/datalab/contrib/mlworkbench).

rajivpb:
Thanks, moved to /contrib now. PTAL.

google/datalab/kernel/__init__.py
```diff
@@ -30,6 +30,7 @@
 # Import the modules that do cell magics.
 import google.datalab.bigquery.commands
 import google.datalab.commands
+import google.datalab.contrib.pipeline.commands
```
qimingj: This line needs to be removed so %%pipeline is not automatically loaded, just as %%ml.

rajivpb:
Thanks, Bradley.

PR #485, opened 2017-08-24
Constraining datetime strings to be in the ISO 8601 format: '%Y-%m-%d… …T%H:%M:%SZ'

rajivpb:
b/64771800


google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -41,16 +41,17 @@ class Pipeline(object):
 from airflow.contrib.operators.bigquery_table_delete_operator import BigQueryTableDeleteOperator
 from airflow.contrib.operators.bigquery_to_bigquery import BigQueryToBigQueryOperator
 from airflow.contrib.operators.bigquery_to_gcs import BigQueryToCloudStorageOperator
-from datetime import datetime, timedelta
+from datetime import timedelta
```
yebrahim: Why not keep importing `datetime` from `datetime` as well?

google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -104,6 +107,12 @@ def py(self):
     return Pipeline._imports + default_args + dag_definition + \
         task_definitions + up_steam_statements
 
+  @staticmethod
+  def _get_datetime_expr(date_string):
```
yebrahim: Should this be called something like `_get_datetime_expr_str`?

google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -41,16 +41,17 @@ class Pipeline(object):
...
 
 """
 
   _default_args_format = """
     'owner': 'Datalab',
     'depends_on_past': False,
     'email': ['{0}'],
-    'start_date': datetime.strptime('{1}', '{3}'),
-    'end_date': datetime.strptime('{2}', '{3}'),
+    'start_date': {1},
```
yebrahim: If this is the case, I'd suggest we change this name to `start_date_utc` and drop the `Z` at the end if it's specified.

tests/kernel/pipeline_tests.py
```diff
@@ -122,14 +121,15 @@ def test_create_cell(self, mock_default_context, mock_notebook_environment):
...
+from datetime import timedelta
+from pytz import timezone
 
 default_args = {
     'owner': 'Datalab',
     'depends_on_past': False,
     'email': ['foo@bar.com'],
-    'start_date': datetime.strptime('Jun 1 2005  1:33PM', '%b %d %Y %I:%M%p'),
-    'end_date': datetime.strptime('Jun 10 2005  1:33PM', '%b %d %Y %I:%M%p'),
+    'start_date': datetime.datetime.strptime('2009-05-05 22:28:15', '%Y-%m-%dT%H:%M:%SZ').replace(tzinfo=timezone('UTC')),
```
yebrahim: Shouldn't the format be `'%Y-%m-%d %H:%M:%S'`? Same below? Or how does this work?

tests/pipeline/pipeline_tests.py
```diff
@@ -102,21 +102,39 @@ def test_get_dag_definition(self):
...
+    self.assertEqual(datetime_expr,
+                     'datetime.datetime.strptime(\'2009-05-05T22:28:15Z\', '
+                     '\'%Y-%m-%dT%H:%M:%SZ\').replace(tzinfo=timezone(\'UTC\'))')
+    self.assertEqual(eval(datetime_expr),
+                     datetime.datetime(2009, 5, 5, 22, 28, 15,
+                                       tzinfo=timezone('UTC')))
+
+    # Incorrectly formatted strings should throw (in this case without the 'Z')
+    with self.assertRaises(ValueError):
+      eval(pipeline.Pipeline._get_datetime_expr('2009-05-05T22:28:15'))
```
yebrahim: Can we drop this `eval`?

google/datalab/contrib/pipeline/_pipeline.py
yebrahim: Also, why don't we evaluate the expression instead of returning a string to be evaluated later?

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: No real reason. I think I was just trying to reduce complexity from the imports and move them down to the invocation sites in cases where the number of invocations are low (2 in this case).

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: These are actually part of the airflow python spec, and cannot be changed. I plan to make it clear in our documentation in Datalab that the allowed timezone is only UTC.

tests/kernel/pipeline_tests.py
rajivpb: The format should have been '%Y-%m-%dT%H:%M:%SZ', and I've changed it now. It didn't actually matter because this code doesn't get executed and is only meant for string comparison for the tests.

tests/pipeline/pipeline_tests.py
rajivpb: This is actually testing that the code generated is valid python (i.e. it executes). It is generated based on user input. In this particular line, we are checking that it is not fact not valid in cases where the user input is not according to the specified format (i.e.  '%Y-%m-%dT%H:%M:%SZ').

google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -41,16 +41,17 @@ class Pipeline(object):
 from airflow.contrib.operators.bigquery_table_delete_operator import BigQueryTableDeleteOperator
 from airflow.contrib.operators.bigquery_to_bigquery import BigQueryToBigQueryOperator
 from airflow.contrib.operators.bigquery_to_gcs import BigQueryToCloudStorageOperator
-from datetime import datetime, timedelta
+from datetime import timedelta
+from pytz import timezone
 
 """
 
   _default_args_format = """
```
jimmc: Consider giving this a more descriptive name that makes it clear that this is generating a set of standard Airflow attributes. Or, less preferable, add a comment.

tests/kernel/pipeline_tests.py
jimmc: Why do you have code in a test that is not getting executed?

jimmc: Looks like you need to add a T and Z to the sample date/time string to match the specified format. Also in end_date here, and on lines 191 and 192.

google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -104,6 +107,12 @@ def py(self):
     return Pipeline._imports + default_args + dag_definition + \
         task_definitions + up_steam_statements
 
+  @staticmethod
+  def _get_datetime_expr_str(date_string):
+    expr_format = 'datetime.datetime.strptime(\'{1}\', \'{2}\').replace(tzinfo=timezone(\'UTC\'))'
+    datetime_format = '%Y-%m-%dT%H:%M:%SZ'  # ISO 8601, always UTC
+    return expr_format.format(date_string, date_string, datetime_format)
```
jimmc: You are passing three arguments to a formatting string that only has two replacement fields. I think you can drop the first date_string here, and change the 1 and 2 in the format string to 0 and 1.

google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -104,6 +107,12 @@ def py(self):
     return Pipeline._imports + default_args + dag_definition + \
         task_definitions + up_steam_statements
 
+  @staticmethod
+  def _get_datetime_expr_str(date_string):
```
jimmc: Why are you returning an expression containing a call to strptime? Can  you instead do the call to strptime here and return the resulting string in quotes?

rajivpb: I had it that way earlier, but the code wouldn't evaluate:

>>> datetime.datetime.strptime('2009-05-05T22:28:15Z', '%Y-%m-%dT%H:%M:%SZ').replace(tzinfo=timezone('UTC'))
datetime.datetime(2009, 5, 5, 22, 28, 15, tzinfo=<UTC>)
>>> datetime.datetime(2009, 5, 5, 22, 28, 15, tzinfo=<UTC>)
  File "<stdin>", line 1
    datetime.datetime(2009, 5, 5, 22, 28, 15, tzinfo=<UTC>)
                                                     ^
SyntaxError: invalid syntax

rajivpb: I tried that earlier, but the resulting string in quotes does not evaluate:
```
>>> datetime.datetime.strptime('2009-05-05T22:28:15Z', '%Y-%m-%dT%H:%M:%SZ').replace(tzinfo=timezone('UTC'))
datetime.datetime(2009, 5, 5, 22, 28, 15, tzinfo=<UTC>)
>>> datetime.datetime(2009, 5, 5, 22, 28, 15, tzinfo=<UTC>)
  File "<stdin>", line 1
    datetime.datetime(2009, 5, 5, 22, 28, 15, tzinfo=<UTC>)
                                                     ^
SyntaxError: invalid syntax

```

rajivpb: Also, changed this a little bit to work-around yaml.load() returning unaware datetime objects.

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: Good catch, thanks! It's a happy accident that this worked, but thinking again that's probably because I had the {0} there at one point and then removed it.

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: Done; refactored and renamed to airflow_default_args.

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: Renamed this now. We can't evaluate the expression because we want the string as-is so that it becomes part of the airflow python spec.

tests/kernel/pipeline_tests.py
rajivpb: Done; I refactored this around a little bit, but net net is that I've added a T and Z now.

google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -104,6 +92,40 @@ def py(self):
...
+
+  @staticmethod
+  def _get_datetime_expr_str(datetime_obj):
+    # User is expected to always provide start_date and end_date in UTC and in
+    # the %Y-%m-%dT%H:%M:%SZ format (i.e. _with_ the trailing 'Z' to
+    # signify UTC).
+    # However, due to a bug/feature in yaml.load(), strings that look like
+    # datetimes are parsed into timezone _unaware_ datetime objects (even if
+    # they do have a timezone). Hence we will be highly restrictive in the
+    # format we accept and will only allow '%Y-%m-%dT%H:%M:%SZ' format.
```
jimmc: This comment says we only allow a format that ends with Z, but the format on the next line does not end with Z. Is one of these incorrect?

tests/kernel/pipeline_tests.py
```diff
@@ -40,9 +40,8 @@ class TestCases(unittest.TestCase):
   sample_cell_body = """
 email: foo@bar.com
 schedule:
-  start_date: Jun 1 2005  1:33PM
-  end_date: Jun 10 2005  1:33PM
-  datetime_format: '%b %d %Y %I:%M%p'
+  start_date: 2009-05-05T22:28:15Z
```
jimmc: These test lines, and some below, include a Z, but I don't see Z used in the format in the main code. Do these tests pass? Do they also pass without the Z, and do they fail if you add some other junk after the Z?

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: tl;dr - both are correct. 

For now, '%Y-%m-%dT%H:%M:%SZ' will be our advertised format of accepting UTC date-times. 

The '%Y-%m-%dT%H:%M:%S' is an implementation detail. I'm using it to specify the datetime, and then am explicitly setting the timezone('UTC') on the datetime (L126).

Eventually, once I implement yaml validation against a schema (https://b.corp.google.com/issues/64951979#comment2), I'll throw an error when users do _not_ specify their string in the '%Y-%m-%dT%H:%M:%SZ' format.

tests/kernel/pipeline_tests.py
rajivpb: The tests pass because yaml.load() will successfully parse 2009-05-05T22:28:15Z. For now, they will pass without the Z as well, but not after https://b.corp.google.com/issues/64951979#comment2 is fixed.

google/datalab/contrib/pipeline/_pipeline.py
jimmc: Perhaps you could add a TODO about what needs to be done and when.

rajivpb:
Jim: I added the TODO like you've suggested. Also just merged master. Looks like that automatically marked your review as stale (I'm still learning the vagaries of git). Looks like your approval is required again. PTAL.

rajivpb:
Thanks a lot Jim!

PR #488, opened 2017-08-31
Introducing support for the bq.load and bq.execute task-types in %pipeline

rajivpb:


google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -182,48 +163,123 @@ def _get_dependency_definition(task_id, dependencies):
...
-          if param_name not in task_details:
-            task_details[param_name] = param_value
+  def _get_bq_execute_params(operator_task_details):
+    if 'query' in operator_task_details:
+      operator_task_details['bql'] = operator_task_details['query'].sql
+      del operator_task_details['query']
+
+    # Add over-rides of Airflow defaults here.
+    if 'use_legacy_sql' not in operator_task_details:
+      operator_task_details['use_legacy_sql'] = False
```
chmeyers: Datalab only supports GoogleSQL for BigQuery, so should force use_legacy_sql to always be false?

google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -182,48 +163,123 @@ def _get_dependency_definition(task_id, dependencies):
...
+
+    return operator_task_details
+
+  @staticmethod
+  def _get_bq_extract_params(operator_task_details):
+    if 'table' in operator_task_details:
+      table = google.datalab.bigquery.commands._bigquery._get_table(operator_task_details['table'])
+      operator_task_details['source_project_dataset_table'] = table.full_name
+      del operator_task_details['table']
```
chmeyers: Are there other members of operator_task_details, or are the 6 we translate here the entire set?

If this is the entire set, it might be cleaner to just create a new dict.  If it's not this is fine.

google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -182,48 +163,123 @@ def _get_dependency_definition(task_id, dependencies):
...
+    if 'table' in operator_task_details:
+      table = google.datalab.bigquery.commands._bigquery._get_table(operator_task_details['table'])
+      if not table:
+        table = google.datalab.bigquery.Table(operator_task_details['table'])
+        # TODO(rajivpb): Ensure that mode == create here.
+      operator_task_details['destination_project_dataset_table'] = table.full_name
+      del operator_task_details['table']
+
+    if 'format' in operator_task_details:
+      operator_task_details['export_format'] = 'CSV' if operator_task_details['format'] == 'csv' \
```
chmeyers: Google style doesn't allow for backslash line continuation and instead requires implicit line joining by wrapping things in parens.  We don't always follow Google style but it's a nice-to-have.

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: Is that a question about why it's false here? Or is that a statement about it needing to be True?

Airflow makes this True by default:
https://github.com/apache/incubator-airflow/blob/master/airflow/contrib/operators/bigquery_operator.py#L68

And Datalab makes this False by default:
https://github.com/googledatalab/pydatalab/blob/b27dc582882fe5497af49d45fc28204e4c46db12/google/datalab/bigquery/_api.py#L170

We need to be consistent with the rest of Datalab here. Hence, we need to override the Airflow default (of True) and set it to False.

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: Acknowledged; will keep in mind and retro-fix opportunistically.

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: There could be others as well and those would need to just get passed through. My thoughts are that we don't want to be stuck each time there is a change (mostly additions, hopefully) to the parameter set in Airflow's operators, and we want a Datalab user to be able to specify it in a cell (unless of course, it needs translation in which we'll make a code change here).

rajivpb:
Chris: I actually added a few unit-tests to increase test coverage (mistakenly believing that this was a merge blocker), and need an approval from you again (ouch). PTAL, and thanks for review comments so far.

rajivpb:
Thanks Chris, Yasser for feedback.

PR #489, opened 2017-08-31
Interface for %bq pipeline

rajivpb:
Just the arguments and the interface. Wanted to get some early feedback on what thoughts were. 

Idea is for this to orchestrate just bq-load, bq-execute, and bq-extract; IOW a fixed and static DAG. Assumption is that this is a common scenario for data-scientists.


google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
...
+  pipeline_parser.add_argument('-nc', '--nocache', help='Don\'t use previously cached results',
+                               action='store_true')
+
+  # load arguments
+  pipeline_parser.add_argument(
+      '-lm', '--load_mode', help='One of create (default), append or overwrite for loading data',
+      choices=['create', 'append', 'overwrite'], default='create')
+  pipeline_parser.add_argument('-lf', '--load_format', help='The format of the source file in GCS',
+                               choices=['json', 'csv'], default='csv')
+  pipeline_parser.add_argument('--skip',
```
chmeyers: --skip_lines?

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
...
+                                                  'the cell body, otherwise the schema is inferred '
+                                                  'from existing table.')
+
+  # common arguments
+  pipeline_parser.add_argument('-b', '--billing', type=int, help='BigQuery billing tier')
+  # TODO(rajivpb): I've naively made this a common param. Any objections?
+  pipeline_parser.add_argument('-nc', '--nocache', help='Don\'t use previously cached results',
+                               action='store_true')
+
+  # load arguments
```
chmeyers: If these argument groups are for different subcommands, should they be using subparsers and be specific to the subcommand?

That way you'd be saying "pipeline --nocache load --mode" instead of "pipeline --nocache --load_mode"

As it is, how do you know it even is a load command?  What if there is a mix of both load and extract flags for example.

google/datalab/bigquery/commands/_bigquery.py
rajivpb: But, load uses just 'skip' and I want to be consistent:
https://github.com/googledatalab/pydatalab/blob/master/google/datalab/bigquery/commands/_bigquery.py#L305

google/datalab/bigquery/commands/_bigquery.py
rajivpb: Idea is for the user to be able to specify:
%%bq pipeline <load args> <execute args> <extract args>

Datalab will then create the necessary operators to execute the fixed DAG of load-execute-extract. So we want a single subcommand with the options of all these three steps, and we want to make it as conceptually similar to what already exists in Datalab as possible.

This is a small divergence from the design of %%pipeline. Nikhil and I talked about this on Friday and he had felt that it might be easier to ship something like this for starters.

rajivpb: And to answer your other question, if the user indeed desires to do both a load and extract, the appropriate load and extract params will need to be specified. There are a few parameters that are common to the bq-load and bq-extract subcommands, so when specifying them inside bq-pipeline, I've prefixed them with 'l/load' or 'x/extract' in order to disambiguate.

rajivpb:
Bradley: I require your approval because I made a few changes as per your previous comments. PTAL.

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
...
+                               help='The number of head lines to skip during load; useful for CSV',
+                               type=int, default=0)
+  pipeline_parser.add_argument('-s', '--strict', help='Whether to reject bad values and jagged '
+                                                      'lines when loading', action='store_true')
+  pipeline_parser.add_argument(
+      '-ld', '--load_delimiter', default=',', help='The inter-field delimiter for CVS (default ,) '
+                                                   'in load file')
+  pipeline_parser.add_argument(
+      '-lq', '--load_quote', default='"', help='The quoted field delimiter for CVS (default ") in '
+                                               'the load file')
```
jimmc: You say "in the load file" in this help, but "in load file" in the previous help (line 345)

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
...
+      '-l', '--large', help='Whether to allow large results during execution', action='store_true')
+  pipeline_parser.add_argument('-eq', '--execute_query', help='The name of query for execution',
+                               required=True)
+  pipeline_parser.add_argument(
+      '-et', '--extract_table', help='Destination table name for the execution results')
+
+  # extract arguments
+  pipeline_parser.add_argument(
+      '-xf', '--extract_format', choices=['csv', 'json'], default='csv',
+      help='The format to use for the export for extraction.')
```
jimmc: This appears to be the only place you use the word 'export'. I would drop it.

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
...
+  pipeline_parser.add_argument(
+      '-c', '--compress', action='store_true',
+      help='Whether or not to compress the data after extraction')
+  pipeline_parser.add_argument(
+      '-H', '--header', action='store_true',
```
jimmc: Why is this called just "header" when the next arg is called "extract_delimiter"? Maybe this should be "extract_header" to make it more in line with the other extract options. Maybe also "extract_compress"?

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
...
+      help='Whether to include a header line (CSV only) in extracted file')
+  pipeline_parser.add_argument(
+      '-xd', '--extract_delimiter', default=',',
+      help='The field delimiter to use (CSV only) for extraction')
+  # TODO(rajivpb): Consider not having this, and defaulting to "SELECT * FROM <composed-table-name>"
```
jimmc: Yeah, I'm confused as to why you would need a second query statement. Seems like the point of the execution step is to create the output that you want, in a temporary table, and the extraction step is just telling how that output should be dumped.

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
   return load_parser
 
 
+def _create_pipeline_subparser(parser):
+  pipeline_parser = parser.subcommand('pipeline', 'Load from GCS, query, and extract data from GCS '
```
nikhilk: I think the command description should be short single liner. Something like:

    Creates a pipeline to execute a SQL query to transform data using BigQuery.

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
...
+                                                  'into a BigQuery table. If creating a new table, '
+                                                  'a schema should be specified in YAML or JSON in '
+                                                  'the cell body, otherwise the schema is inferred '
+                                                  'from existing table.')
+
+  # common arguments
+  pipeline_parser.add_argument('-b', '--billing', type=int, help='BigQuery billing tier')
+  # TODO(rajivpb): I've naively made this a common param. Any objections?
+  pipeline_parser.add_argument('-nc', '--nocache', help='Don\'t use previously cached results',
```
nikhilk: Does this make sense to specify in a pipeline? Even if it does, no-cache should be true by default in a pipeline, so a pipeline doesn't pick up state from some other run of some random other query.

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
...
+                               action='store_true')
+
+  # load arguments
+  pipeline_parser.add_argument(
+      '-lm', '--load_mode', help='One of create (default), append or overwrite for loading data',
```
nikhilk: I think we want the concept of an input data source which can either be an existing table or a set of files in GCS. This is better represented as a section in the YAML spec rather than a set of arguments. Example:

    input:
      type: table|gcs
      path: gs://... (with placeholders for variables)
      name: table-name-with-placeholders-for-date-or-some-other-variable

Also, I don't know if we should _actually_ load data into a table. We could probably just use a federated data source and query directly against the GCS data. Otherwise it raises questions of who/what deletes the table after the pipeline has run.

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
...
+  pipeline_parser.add_argument('-lt', '--load_table', help='The destination bigquery table name '
+                                                           'for loading.')
+
+  # execute arguments
+  pipeline_parser.add_argument(
+      '-em', '--execute_mode', help='The table creation mode for execution', default='create',
+      choices=['create', 'append', 'overwrite'])
+  pipeline_parser.add_argument(
+      '-l', '--large', help='Whether to allow large results during execution', action='store_true')
+  pipeline_parser.add_argument('-eq', '--execute_query', help='The name of query for execution',
```
nikhilk: Just `query`. Since there is only one query in the picture, it likely doesn't need to be qualified.

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
...
+                               required=True)
+  pipeline_parser.add_argument(
+      '-et', '--extract_table', help='Destination table name for the execution results')
```
nikhilk: Similar to input, we should have an output section in the YAML that describes the output -- table or gcs, with similar support for placeholders to be able to generate per pipeline run outputs.

google/datalab/bigquery/commands/_bigquery.py
rajivpb: Understood. Removed this in https://github.com/googledatalab/pydatalab/pull/494 and have hard-coded this to False. At the time of coding this up, I hadn't completely dived into what this parameter meant, but it makes sense now.

google/datalab/bigquery/commands/_bigquery.py
rajivpb: Fixed in https://github.com/googledatalab/pydatalab/pull/494

google/datalab/bigquery/commands/_bigquery.py
rajivpb: Fixed in https://github.com/googledatalab/pydatalab/pull/494

google/datalab/bigquery/commands/_bigquery.py
rajivpb: Removed in https://github.com/googledatalab/pydatalab/pull/494

google/datalab/bigquery/commands/_bigquery.py
rajivpb: I prefixed only those arguments that are common to another of {load | execute | extract} to disambiguate.

google/datalab/bigquery/commands/_bigquery.py
rajivpb: Dropped in https://github.com/googledatalab/pydatalab/pull/494

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
...
+  pipeline_parser.add_argument('-lp', '--load_path', help='The path URL of the GCS load file(s)')
+  pipeline_parser.add_argument('-lt', '--load_table', help='The destination bigquery table name '
+                                                           'for loading.')
+
+  # execute arguments
+  pipeline_parser.add_argument(
+      '-em', '--execute_mode', help='The table creation mode for execution', default='create',
+      choices=['create', 'append', 'overwrite'])
+  pipeline_parser.add_argument(
+      '-l', '--large', help='Whether to allow large results during execution', action='store_true')
```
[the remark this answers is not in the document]
rajivpb: Done in https://github.com/googledatalab/pydatalab/pull/494

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -316,6 +316,73 @@ def _create_load_subparser(parser):
...
+  pipeline_parser.add_argument(
+      '-lq', '--load_quote', default='"', help='The quoted field delimiter for CVS (default ") in '
+                                               'the load file')
+  pipeline_parser.add_argument('-lp', '--load_path', help='The path URL of the GCS load file(s)')
+  pipeline_parser.add_argument('-lt', '--load_table', help='The destination bigquery table name '
+                                                           'for loading.')
+
+  # execute arguments
+  pipeline_parser.add_argument(
+      '-em', '--execute_mode', help='The table creation mode for execution', default='create',
```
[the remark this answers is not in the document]
rajivpb: Now that they are in the comments, I've enumerated them uniformly in https://github.com/googledatalab/pydatalab/pull/494

google/datalab/bigquery/commands/_bigquery.py
rajivpb: Changed this to consistent wording (i.e. always using the definite article) in https://github.com/googledatalab/pydatalab/pull/494

google/datalab/bigquery/commands/_bigquery.py
rajivpb: Agreed on this being in the YAML. I already changed this in https://github.com/googledatalab/pydatalab/pull/494

I have been heretofore thinking of a user experience that specified a pipeline in a way that was conceptually familiar to a Datalab user - i.e. having it looked like it is composed of bq.load, bq.execute, and bq.extract. But I see the simplicity and compactness in your proposed user experience, so let me flesh this out some more with the other required parameters for such an experience (and also respond more fully to your comments in https://github.com/googledatalab/pydatalab/pull/494). 

Perhaps we can leave the composition of individual bq tasks to %%pipeline.

google/datalab/bigquery/commands/_bigquery.py
rajivpb: Yeah, see response above for your comment on L334.

google/datalab/bigquery/commands/_bigquery.py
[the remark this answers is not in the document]
rajivpb: Turns out that all of these are now part of the config in yaml and not command-line arguments. Doesn't mean that they don't get documented anywhere of course, but just that the location is TBD. For now, I've provisionally added them in comments against where they first get used and have incorporated your feedback (in PR https://github.com/googledatalab/pydatalab/pull/494). Once the documentation location gets finalized, I'll move them there.

PR #495, opened 2017-09-02
Add "%%ml explain" to mlworkbench.

qimingj:
- API for explainer
- magic

Also, include the change to make "headers" and "image_columns" unneeded in prediction with local models. Instead we get headers and image_columns from local model's assets files. In the future, we can do same for online models.

There is also an ongoing change to expose inception checkpoint as transform config item, so we don't have to download from GCS and can enable the image tests in travis.

### DIFF
--- a/google/datalab/contrib/mlworkbench/_prediction_explainer.py
+++ b/google/datalab/contrib/mlworkbench/_prediction_explainer.py
@@ -41,10 +41,10 @@ class PredictionExplainer(object):
         self._headers = [x['name'] for x in schema]
         self._text_columns, self._image_columns = [], []
         for k, v in six.iteritems(features):
-          if v['transform'] in ['image_to_vec']:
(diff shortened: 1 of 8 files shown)


google/datalab/contrib/mlworkbench/_prediction_explainer.py
```diff
@@ -0,0 +1,179 @@
...
+        Args:
+          model_dir: the directory of the model to use for prediction.
+        """
+
+        self._model_dir = model_dir
+        schema, features = _local_predict.get_model_schema_and_features(model_dir)
+        self._headers = [x['name'] for x in schema]
+        self._text_columns, self._image_columns = [], []
+        for k, v in six.iteritems(features):
+          if v['transform'] in ['image_to_vec']:
```
chmeyers: Your indentation in this file is inconsistent.

PR #509, opened 2017-09-08
Fix a few %%ml issues

qimingj:
- project-id should be optional when running %%ml transform --cloud.
- %%ml batch_predict --cloud supports models in gcs (not deployed).


qimingj:
Chris, PTAL. One small change.

PR #511, opened 2017-09-09
Fix two %%bq load/extract issues.

qimingj:
- %%bq load with json fails with "Invalid source format NEWLINE_DELIMITED_JSON"
- %%bq extract with json fails because we set a csv_delimiter and BQ doesn't like it.


qimingj:
Hold off review. Need to fix test.

PR #483, opened 2017-08-22
Bumping google.cloud from 0.19 to 0.27

jpvelez:
Hi there,

Having similar version collision problems as in the previous PR. Was hoping we could bump `google.cloud` dependency from 0.19 to 0.27

Issue #421, opened 2017-06-20
Migrate to google-auth


parthea:
[oauth2client is deprecated](https://github.com/google/oauth2client/releases/tag/v4.1.0) as of May 10th. No more features will be added to the libraries and the core team is turning down support. [google-auth](https://google-auth.readthedocs.io/) is recommended.

yebrahim:
Fixed by #516.

PR #516, opened 2017-09-11
Switching to google.auth

yebrahim:
Following up on https://github.com/googledatalab/pydatalab/pull/483 and https://github.com/googledatalab/pydatalab/issues/421, this PR moves away from `oauth2client` to `google.auth`.


datalab/context/_utils.py
```diff
@@ -75,9 +79,10 @@ def get_credentials():
       # Use the first gcloud one we find
       for entry in creds['data']:
         if entry['key']['type'] == 'google-cloud-sdk':
-          return oauth2client.client.OAuth2Credentials.from_json(json.dumps(entry['credential']))
+          creds = oauth2client.client.OAuth2Credentials.from_json(json.dumps(entry['credential']))
+          return google.auth._oauth2client.convert(creds)
```
qimingj: Might be worthwhile to just create a new credential out of the old one using the impl (https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/master/google/auth/_oauth2client.py#L58) of the converter, in case they rename/move the file since it is not a public API yet. I guess it's Okay to use the internal converter too since we pin the version. Up to you.

yebrahim: Fair point. Done.

PR #520, opened 2017-09-12
Switch to using TensorFlow 1.3 in tests.

craigcitro:
Everything *seems* to pass, which is almost too good to be true. 😉 

I didn't touch the explicit paths in `mltoolbox`, but let me know if you'd like those to stay in sync.


craigcitro:
PTAL @chmeyers @qimingj (forgot I couldn't add reviewers)

craigcitro:
As per offline convo, dropping this since Datalab is skipping 1.3.

PR #532, opened 2017-09-27
Have Tensorboard respect basepath of Datalab.

qimingj:


google/datalab/ml/_tensorboard.py
```diff
@@ -26,13 +28,25 @@
...
 class TensorBoard(object):
-  """Start, shutdown, and list TensorBoard instances.
-  """
+  """Start, shutdown, and list TensorBoard instances."""
+
+  @staticmethod
+  def _get_base_path():
+    """Get base path from environment variable. If not exists, returns ''."""
+
+    if 'DATALAB_SETTINGS_OVERRIDES' not in os.environ:
```
yebrahim: I think if you take out this `if` it's cleaner. Any issues with getting the env var below will just end up returning an empty string.

yebrahim:
I'll merge this so as not to block progress on this bug.

PR #536, opened 2017-09-29
Add a google.datalab.ml.Metrics class, which computes accuracy, RMSE, MAE, Percentile from raw evaluation data.

qimingj:
…

### DIFF
--- a/google/datalab/ml/_metrics.py
+++ b/google/datalab/ml/_metrics.py
@@ -88,7 +88,7 @@ class Metrics(object):
       a Metrics instance.
 
     Args:
-      bigquery: A BigQuery table name or a query.
+      sql: A BigQuery table name or a query.
     """
 
     if isinstance(sql, bq.Query):
@@ -126,7 +126,7 @@ class Metrics(object):
     """Get accuracy numbers for each target and overall.
 
     Returns:
-      A DataFrame with two columns: 'class' and 'accuracy'. It contains the overall
+      A DataFrame with two columns: 'class' and 'accuracy'. It also contains the overall
       accuracy with class being '_all'.
 
     Raises:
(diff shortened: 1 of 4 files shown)


google/datalab/ml/_metrics.py
```diff
@@ -0,0 +1,268 @@
...
+        correct_count = len(df[(df['target'] == df['predicted']) & (df['target'] == label)])
+        total_count = len(df[(df['target'] == label)])
+        accuracy_results.append({
+            'target': label,
+            'accuracy': float(correct_count) / total_count,
+            'count': total_count
+        })
+
+      total_correct_count = len(df[(df['target'] == df['predicted'])])
+      total_accuracy = float(total_correct_count) / len(df)
```
chmeyers: Just double checking, is len(df) ever zero?

google/datalab/ml/_metrics.py
```diff
@@ -0,0 +1,268 @@
...
+      number.
+    """
+
+    if self._input_csv_files:
+      df = self._get_data_from_csv_files()
+      if 'target' not in df or 'predicted' not in df:
+        raise ValueError('Cannot find "target" or "predicted" column')
+
+      df = df[['target', 'predicted']].apply(pd.to_numeric)
+      mse = mean_squared_error(df['target'], df['predicted'])
```
chmeyers: Is mse ever negative or a NaN?

google/datalab/ml/_metrics.py
qimingj: Done. Also did similar check in a few other places, and tested 0 instance in csv and 0 instance in bq.

google/datalab/ml/_metrics.py
qimingj: Tested it. If NaN, sklearn raises understandable error.

qimingj:
PTAL.

Issue #540, opened 2017-10-04
parameters should be specifiable in a friendlier format


rajivpb:
Current: 
parameters:
  - name: endpoint
    type: STRING
    value: Interact2
  - name: table_format
    type: STRING
    value: $table_format

Desired:
  parameters:
  endpoint: Interact2
  table_format: $table_format

The type of the parameter should be generally inferable from the parameter value and it would be interesting to see if that holds true for the range of parameter types supported in BQ jobs (though this needs to be confirmed).

Issue #505, opened 2017-09-07
Expose common airflow macros in pydatalab's pipeline magics/subcommand via implicit parameters


rajivpb:

rajivpb:
https://github.com/googledatalab/pydatalab/pull/554

PR #542, opened 2017-10-05
Introducing support for Airflow macros in Pipeline()

rajivpb:


rajivpb:
This is a fix for https://github.com/googledatalab/pydatalab/issues/505

google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -29,13 +29,25 @@ class Pipeline(object):
...
-from google.datalab.contrib.bigquery.operators.bq_execute_operator import ExecuteOperator
-from google.datalab.contrib.bigquery.operators.bq_extract_operator import ExtractOperator
+from google.datalab.contrib.bigquery.operators._bq_load_operator import LoadOperator
+from google.datalab.contrib.bigquery.operators._bq_execute_operator import ExecuteOperator
+from google.datalab.contrib.bigquery.operators._bq_extract_operator import ExtractOperator
 from datetime import timedelta
 from pytz import timezone
 """
 
+  _airflow_macros = {
```
nikhilk: Comments would be useful (perhaps with examples).

Also
- I wonder if all those are required (the nodash options, and yesterday/tomorrow)?
- What is the difference between ds and execution_date?

I am hoping it could be as simple as execution_timestamp and execution_date. Seeing examples would help be more definitive.

tests/bigquery/pipeline_tests.py
```diff
@@ -116,57 +114,114 @@ def test_get_extract_parameters(self):
...
+        'SELECT @column FROM publicdata.samples.wikipedia where endpoint=@endpoint')
+
+    transformation_config = {
+      'query': 'foo_query'
+    }
+    output_config = {
+      'table': 'foo_table',
+      'mode': 'foo_mode'
+    }
+    parameters_config = [
```
nikhilk: Can we simplify these? Or in another change?

rajivpb:
Cloning this in https://github.com/googledatalab/pydatalab/pull/550 because this is based off of a branch that has been deleted. Will address all comments in the new PR.

tests/bigquery/pipeline_tests.py
rajivpb: Different change. I have a bug on it already: https://github.com/googledatalab/pydatalab/issues/540

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: Done. I've reduced it to just the two - ds and ts. I think there may be value in adding the following as well though:

    'yesterday_ds': '{{yesterday_ds}}',  # yesterday’s date as YYYY-MM-DD
    'tomorrow_ds': '{{tomorrow_ds}}',  # tomorrow’s date as YYYY-MM-DD

PR #547, opened 2017-10-10
Making email an optional/customizable field in cell_body

rajivpb:
This had been provisionally hard-coded to foo@bar.com in various places. This now being surfaced as a legit settable key in the config for the pipeline in the cell-body.

### DIFF
--- a/tests/kernel/bigquery_tests.py
+++ b/tests/kernel/bigquery_tests.py
@@ -757,13 +757,7 @@ WITH q1 AS (
 
     @mock.patch('google.datalab.Context.default')
     @mock.patch('google.datalab.utils.commands.notebook_environment')
-    @mock.patch('google.datalab.bigquery.Table.exists')
-    @mock.patch('google.datalab.bigquery.commands._bigquery._get_table')
-    def test_pipeline_cell(self, mock_get_table, mock_table_exists, mock_environment,
-                           mock_default_context):
-      table = bq.Table('project.test.table')
-      mock_get_table.return_value = table
-      mock_table_exists.return_value = True
+    def test_pipeline_cell(self, mock_environment, mock_default_context):
       context = TestCases._create_context()
       mock_default_context.return_value = context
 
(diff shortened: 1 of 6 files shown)


tests/pipeline/pipeline_tests.py
```diff
@@ -309,26 +310,19 @@ def test_get_datetime_expr(self):
...
-}
+    actual = pipeline.Pipeline._get_default_args(dag_dict.get('schedule').get('start'),
+                                                 dag_dict.get('schedule').get('end'),
+                                                 dag_dict.get('email'))
+    self.assertIn("'email': ['foo1@test.com', 'foo2@test.com']", actual)
 
-"""  # noqa
-    )
+    # email is not specified
+    dag_dict = yaml.load(PipelineTest._test_pipeline_yaml_spec)
```
yebrahim: Do we need this second declaration? If it's meant as a separate test case entirely then perhaps move to a different test?

tests/kernel/bigquery_tests.py
```diff
@@ -773,82 +766,24 @@ def test_pipeline_cell(self, mock_get_table, mock_table_exists, mock_environment
...
-                        mode: required
-                        description: description1
               transformation:
                   query: foo_query
               output:
-                  path: test/path
                   table: project.test.table
          """
 
-      output = google.datalab.contrib.bigquery.commands._bigquery._pipeline_cell(args, cell_body)
```
yebrahim: It's not clear to me what these changes to the test are. Is it not testing the pipeline anymore?

tests/pipeline/pipeline_tests.py
rajivpb: Split the tests now.

tests/kernel/bigquery_tests.py
rajivpb: I'm trying to limit the use of the long "golden" file string to as few tests as possible (ideally just two - one for %bq and one for %pipeline. This was a test that used it unnecessarily and had substantial overlap with those tests, so I did some refactoring and shortened this.

Issue #501, opened 2017-09-07
%pipeline 'create' subcommand should copy over airflow python spec to go/encom


rajivpb:

rajivpb:
https://github.com/googledatalab/pydatalab/pull/548

PR #548, opened 2017-10-10
Copy airflow python spec to GCS

rajivpb:
The bucket name that Composer expects is hardcoded for now; will be fixed as part of https://github.com/googledatalab/pydatalab/issues/501

This PR just has the helper routines for this.


google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -84,8 +69,35 @@ def get_airflow_spec(self):
...
+    bucket = client.get_bucket('airflow-staging-test36490808-bucket')
+    filename = 'dags/{0}.py'.format(self._name)
+    blob = storage.blob.Blob(filename, bucket)
+    blob.upload_from_string(self._get_airflow_spec())
+
+  @staticmethod
+  def get_pipeline_spec(spec_str, env=None):
+    """ Gets a dict representation of the pipeline-spec, given a yaml string.
+    Args:
+      name: name of the pipeline (line argument) from Datalab
```
qimingj: arg name mismatch?

rajivpb: Fixed.

tests/pipeline/pipeline_tests.py
```diff
@@ -51,7 +51,7 @@ class PipelineTest(unittest.TestCase):
 
   @staticmethod
   def _create_context():
-    project_id = 'test'
+    project_id = 'test.project'
```
jimmc: Read project IDs can not contain a period, only lowercase letters, digits, or hyphens.

rajivpb: Removed; thanks!

PR #550, opened 2017-10-11
Introducing support for Airflow macros in Pipeline()

rajivpb:
This is a clone of https://github.com/googledatalab/pydatalab/pull/542/, but based off of master.


google/datalab/_context.py
```diff
@@ -82,6 +82,16 @@ def set_config(self, config):
     """ Set the config property for the context. """
     self._config = config
 
+  def override(self, project=None, zone=None, credentials=None, config=None):
```
nikhilk: Was this supposed to come back?

google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -30,13 +30,19 @@ class Pipeline(object):
...
+from google.datalab.contrib.bigquery.operators._bq_execute_operator import ExecuteOperator
+from google.datalab.contrib.bigquery.operators._bq_extract_operator import ExtractOperator
 from datetime import timedelta
 from pytz import timezone
 """
 
+  # These are documented here: https://airflow.incubator.apache.org/code.html?highlight=macros#default-variables  # noqa
+  _airflow_macros = {
+    'ds': '{{ ds }}',  # the execution date as YYYY-MM-DD
+    'ts': '{{ ts }}',  # same as execution_date.isoformat()
```
nikhilk: Can you just put an example, without referring to execution_date which isn't one of the listed macros?

google/datalab/contrib/bigquery/commands/_bigquery.py
```diff
@@ -147,22 +149,19 @@ def _get_execute_parameters(load_task_config_name, bq_pipeline_transformation_co
     # The name of query for execution; if absent, we return None as we assume that there is
     # no query to execute
     if 'query' not in bq_pipeline_transformation_config:
-      if any(key in bq_pipeline_transformation_config for key in ['large', 'mode']):
-        raise Exception('Query is not specified, but at least one query option is.')
+      # TODO(rajivpb): Log this: 'Query is not specified, but at least one query option is.')
```
yebrahim: Why log this if the check for query options is removed?

google/datalab/contrib/bigquery/commands/_bigquery.py
```diff
@@ -147,22 +149,19 @@ def _get_execute_parameters(load_task_config_name, bq_pipeline_transformation_co
...
-    execute_task_config['query'] = bq_pipeline_transformation_config['query']
-
-    # TODO(rajivpb): There is no unit-test coverage of this.
-
-    # One of 'create' (default), 'append' or 'overwrite' for the destination table in BigQuery
-    execute_task_config['mode'] = bq_pipeline_transformation_config.get('mode', 'create')
-
+    query = utils.commands.get_notebook_item(bq_pipeline_transformation_config['query'])
+    execute_task_config['sql'] = query.sql
+    if 'table' in bq_pipeline_output_config:
```
yebrahim: nit: If you have a default value for these keys, you could do `bq_pipeline_output_config.get('table', None)` for example, instead of the checks.

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: Understood the confusion. Have added more descriptive comments with the full format; hopefully this will address it. Also, I've diverged from airflow's macros by internally mapping 'ds' and 'ts' to 'ds_nodash' and 'ts_nodash'. Since users are likely to use this in file names, I felt that this was safer to do. Additionally, in the spirit of providing an abstraction that was layered on top / simpler than Airflow, I didn't feel compelled to match Airflow's macros 1:1.

google/datalab/contrib/bigquery/commands/_bigquery.py
rajivpb: Removed the TODO now; changed my mind about needing a log statement here since if the user does not specify a 'query' key in the 'transformation' section, it's pretty clear that no transformation will be performed.

google/datalab/contrib/bigquery/commands/_bigquery.py
rajivpb: If the table is not present, I actually do not want to set the key on execute_task_config, hence this check.

google/datalab/_context.py
rajivpb: Nope; there was a bad merge. Removed it now.

rajivpb:
This has been cloned in https://github.com/googledatalab/pydatalab/pull/554, so closing.

PR #556, opened 2017-10-13
ML Workbench Updates

qimingj:
- Introduce "%%ml dataset". Users create datasets and use them for training later. Users can explore these datasets.

- Remove "%%ml predict --cloud" mode.

### DIFF
--- a/google/datalab/ml/_fasets.py
+++ b/google/datalab/ml/_fasets.py
@@ -65,7 +65,7 @@ class FacetsOverview(object):
 class FacetsDiveview(object):
   """Represents A facets overview. """
 
-  def plot(self, data, height=1000):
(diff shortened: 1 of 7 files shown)


google/datalab/ml/_fasets.py
```diff
@@ -65,7 +65,7 @@ def plot(self, data):
 class FacetsDiveview(object):
```
chmeyers: Just realized that this file is named "fasets".  Don't need to actually fix now...

qimingj: Ah. Will fix in a future change.

qimingj:
Fixed a few tests. PTAL.

PR #561, opened 2017-10-19
A few improvements in ML Workbench.

qimingj:
1. Support prediction with image url directly, thus we don't require users to embed image content into CSVs for prediction/batch_prediction.

2. training_data --> data renaming in various %%ml commands.

3. Allow different size of confusion matrix.

### DIFF
--- a/solutionbox/code_free_ml/test_mltoolbox/test_feature_transforms.py
+++ b/solutionbox/code_free_ml/test_mltoolbox/test_feature_transforms.py
@@ -263,7 +263,7 @@ class TestGraphBuilding(unittest.TestCase):
   @unittest.skipIf(not HAS_CREDENTIALS, 'GCS access missing')
   def test_make_transform_graph_images(self):
 
-    print('Testing make_transform_graph with image_to_vec.' +
+    print('Testing make_transform_graph with image_to_vec. ' +
           'It may take a few minutes because it needs to download a large inception checkpoint.')
 
     def _open_and_encode_image(img_url):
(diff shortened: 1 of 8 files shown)


solutionbox/code_free_ml/test_mltoolbox/test_feature_transforms.py
```diff
@@ -282,17 +282,22 @@ def _open_and_encode_image(img_url):
...
       features = {'img': {'transform': 'image_to_vec', 'source_column': 'img'}}
 
+      # Test transformation with encoded image content.
       img_string1 = _open_and_encode_image(
           'gs://cloud-ml-data/img/flower_photos/daisy/15207766_fc2f1d692c_n.jpg')
       img_string2 = _open_and_encode_image(
           'gs://cloud-ml-data/img/flower_photos/dandelion/8980164828_04fbf64f79_n.jpg')
-      input_data = [img_string1, img_string2]
+      # Test transformation with direct file path.
+      img_string3 = 'gs://cloud-ml-data/img/flower_photos/daisy/15207766_fc2f1d692c_n.jpg'
```
chmeyers: Just double checking: These gcs paths are publicly accessible?

qimingj: Yes, although we should probably generate images instead of using ones on GCS. I'll fix that in a separate change.

qimingj:
PTAL.

solutionbox/code_free_ml/test_mltoolbox/test_transform.py
```diff
@@ -46,13 +46,13 @@ def setUpClass(cls):
...
-    image1 = Image.new('RGBA', size=(300, 300), color=(155, 0, 0))
+    image1 = Image.new('RGB', size=(300, 300), color=(155, 0, 0))
     image1.save(img1_file)
     img2_file = os.path.join(cls.source_dir, 'img2.jpg')
-    image2 = Image.new('RGBA', size=(50, 50), color=(125, 240, 0))
+    image2 = Image.new('RGB', size=(50, 50), color=(125, 240, 0))
     image2.save(img2_file)
     img3_file = os.path.join(cls.source_dir, 'img3.jpg')
-    image3 = Image.new('RGBA', size=(800, 600), color=(33, 55, 77))
+    image3 = Image.new('RGB', size=(800, 600), color=(33, 55, 77))
```
yebrahim: Add a test without explicit size to make sure default is used? I'm generally worried about defaults.

qimingj: Size is a required parameter. No default story here.

Issue #506, opened 2017-09-07
Implement XCom for ExecuteOperator to return the table_name


rajivpb:

rajivpb:
Making this P2 since this is not required for '%bq pipeline' or '%ml pipeline'. The DAG and the messages are fixed in these simplified cases, and inter-task message passing is not required.

rajivpb:
Bumping this to P1 because it's required for enabling a P1 scenario - https://github.com/googledatalab/pydatalab/issues/563

rajivpb:
Fixed by https://github.com/googledatalab/pydatalab/pull/577

Issue #499, opened 2017-09-07
pipeline yaml configs should be validated


rajivpb:
(Moving over Datalab issues from buganizer into here. This comes from b/64951979)

This is also being used to track yaml configs for pipeline subcommands (e.g. %bq pipeline, %ml pipeline, etc.)

As part of this, it should also validate that start_date and end_date are specified in UTC (with the 'Z' at the end), i.e. in this format: %Y-%m-%dT%H:%M:%SZ

rajivpb:
Nikhil's library for generating help and validating schema would be useful here.

PR #566, opened 2017-10-21

rajivpb:
This enables the user to leave out the table name in the output section, so the following will now work in a Datalab notebook (in the cloud-ml-dev project):

import google.datalab.kernel
import google.datalab.contrib.pipeline._pipeline

%%bq query --name demo_query
SELECT col1, col2 FROM `cloud-ml-dev.rajivpb_demo.foo_table_created`

%%bq pipeline --name bq_pipeline_demo_20171013
schedule:
  start: 2017-10-20T21:19:15Z
  end: 2017-10-20T22:28:15Z
  interval: '* * * * *'
input:
  path: gs://cloud-ml-dev-nikhilko/iris/data/data_predict.csv
  table: cloud-ml-dev.rajivpb_demo.foo_table_created
  format: csv
  schema:
    - name: prediction
      type: INTEGER
    - name: col1
      type: FLOAT
    - name: col2
      type: FLOAT
    - name: col3
      type: FLOAT
    - name: col4
      type: FLOAT
transformation:
  query: demo_query
output:
  path: gs://rajivpb-airflow-testing/data_predict_query_results_%(ds)s.csv
  format: csv

### DIFF
--- a/google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
+++ b/google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
@@ -31,14 +31,14 @@ class ExecuteOperator(BaseOperator):
 
   def execute(self, context):
     query = bq.Query(sql=self._sql)
-    # allow_large_results is set to False because setting it True requires the destination table to
-    # be specified.
-    # use_cache is set to False since this is most likely the case for all pipeline scenarios
+    # allow_large_results is False because setting it True requires the destination table
+    # use_cache is False since this is most likely the case in pipeline scenarios
     output_options = bq.QueryOutput.table(name=self._table, mode=self._mode, use_cache=False,
                                           allow_large_results=False)
+
     pydatalab_context = google.datalab.Context.default()
     query_params = Pipeline._get_query_parameters(self._parameters)
     job = query.execute(output_options, context=pydatalab_context, query_params=query_params)
-    # returning the name of the table used for the results so that it could be used by downstream
-    # tasks if they need it.
+
+    # Returning the table-name here makes it available for downstream task instances.
     return job.result().name
--- a/google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
+++ b/google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
@@ -10,7 +10,6 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 
-import google
 import google.datalab.bigquery as bq
 from airflow.models import BaseOperator
 from airflow.utils.decorators import apply_defaults
--- a/google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
+++ b/google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
@@ -31,14 +31,15 @@ class ExecuteOperator(BaseOperator):
 
   def execute(self, context):
     query = bq.Query(sql=self._sql)
-    # allow_large_results is False because setting it True requires the destination table
     # use_cache is False since this is most likely the case in pipeline scenarios
+    # allow_large_results can be True only if table is specified (i.e. when it's not None)
     output_options = bq.QueryOutput.table(name=self._table, mode=self._mode, use_cache=False,
-                                          allow_large_results=False)
+                                          allow_large_results=self._table is not None)
 
-    pydatalab_context = google.datalab.Context.default()
     query_params = Pipeline._get_query_parameters(self._parameters)
-    job = query.execute(output_options, context=pydatalab_context, query_params=query_params)
+    job = query.execute(output_options, context=None, query_params=query_params)
 
     # Returning the table-name here makes it available for downstream task instances.
-    return job.result().name
+    return {
+      'table': job.result().name
+    }
--- a/google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
+++ b/google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
@@ -38,6 +38,7 @@ class ExecuteOperator(BaseOperator):
                                           allow_large_results=False)
     pydatalab_context = google.datalab.Context.default()
     query_params = Pipeline._get_query_parameters(self._parameters)
-    # TODO(rajivpb): Is this a sync or an async operation? Unlike load(), this does not wrap its
-    # async counterpart with a job.wait(). Test and wait if necessary.
-    query.execute(output_options, context=pydatalab_context, query_params=query_params)
+    job = query.execute(output_options, context=pydatalab_context, query_params=query_params)
+    # returning the name of the table used for the results so that it could be used by downstream
+    # tasks if they need it.
+    return job.result().name
--- a/google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
+++ b/google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
@@ -36,7 +36,7 @@ class ExecuteOperator(BaseOperator):
                                           allow_large_results=self._table is not None)
 
     query_params = Pipeline._get_query_parameters(self._parameters)
-    job = query.execute(output_options, context=None, query_params=query_params)
+    job = query.execute(output_options, query_params=query_params)
 
     # Returning the table-name here makes it available for downstream task instances.
     return {
--- a/google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
+++ b/google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
@@ -39,13 +39,10 @@ class ExtractOperator(BaseOperator):
       # The task id of the execute task is created by concatenating 'bq_pipeline_execute_task'
       # and '_id'. This is currently being hard-coded, but consider making this a parameter.
       # It could require substantial changes to the underlying object model of Pipeline.
-      self._table = task_instance.xcom_pull(task_ids='bq_pipeline_execute_task_id')
+      execute_task_output = task_instance.xcom_pull(task_ids='bq_pipeline_execute_task_id')
+      self._table = execute_task_output.get('table')
 
-    if not self._table:
-      raise Exception('A table is needed to extract.')
-
-    pydatalab_context = google.datalab.Context.default()
-    source_table = google.datalab.bigquery.Table(self._table, context=pydatalab_context)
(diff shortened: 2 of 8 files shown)


rajivpb:
This should fix https://github.com/googledatalab/pydatalab/issues/506

google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
```diff
@@ -31,13 +31,14 @@ def __init__(self, sql, parameters=None, table=None, mode=None, *args, **kwargs)
 
   def execute(self, context):
     query = bq.Query(sql=self._sql)
-    # allow_large_results is set to False because setting it True requires the destination table to
-    # be specified.
-    # use_cache is set to False since this is most likely the case for all pipeline scenarios
+    # allow_large_results is False because setting it True requires the destination table
+    # use_cache is False since this is most likely the case in pipeline scenarios
     output_options = bq.QueryOutput.table(name=self._table, mode=self._mode, use_cache=False,
                                           allow_large_results=False)
```
nikhilk: Seems like in this case, you are actually producing results into a destination table, so allow_large_results should be True.

google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
```diff
@@ -31,13 +31,14 @@ def __init__(self, sql, parameters=None, table=None, mode=None, *args, **kwargs)
...
+
     pydatalab_context = google.datalab.Context.default()
     query_params = Pipeline._get_query_parameters(self._parameters)
-    # TODO(rajivpb): Is this a sync or an async operation? Unlike load(), this does not wrap its
-    # async counterpart with a job.wait(). Test and wait if necessary.
-    query.execute(output_options, context=pydatalab_context, query_params=query_params)
+    job = query.execute(output_options, context=pydatalab_context, query_params=query_params)
```
nikhilk: Why do you need to pass in the default context explicitly?

google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
```diff
@@ -20,23 +20,36 @@ class ExtractOperator(BaseOperator):
...
+      # If the table is not specified, we fetch it from the output of the execute task, i.e. the
+      # query results table. This could either be a permanent table or a temporary table. If we're
+      # here, it is most likely a temporary table. If it was a permanent one, it would have been
+      # passed in as a param and we wouldn't be here.
+      # TODO(rajivpb): Assert that if we're here, then the table is a temporary one.
+      # TODO(rajivpb):
+      # The task id of the execute task is created by concatenating 'bq_pipeline_execute_task'
+      # and '_id'. This is currently being hard-coded, but consider making this a parameter.
+      # It could require substantial changes to the underlying object model of Pipeline.
+      self._table = task_instance.xcom_pull(task_ids='bq_pipeline_execute_task_id')
```
nikhilk: We should have a consistent model for all operators to produce a dictionary of outputs, so as to support operators that have 0, 1, N outputs, as well as evolve over time to introduce additional outputs.

So it would be good if the execute operator produced an output { 'table': table-name }.


Similarly, I'd expect load and extract to also produce results.

google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
```diff
@@ -20,23 +20,36 @@ class ExtractOperator(BaseOperator):
...
+
+    if not self._table:
+      raise Exception('A table is needed to extract.')
```
nikhilk: What does it mean to get here?

- A mis-configured pipeline? If so, why not try to catch this at pipeline definition time?
- A failure to have an output table during execute? If so, why does the pipeline execution get to this stage? I would expect execute to fail, and for Airflow to retry that task.

google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
```diff
@@ -20,23 +20,36 @@ class ExtractOperator(BaseOperator):
...
+
+    pydatalab_context = google.datalab.Context.default()
```
nikhilk: Similar comment about explicitly passing the default context. It should happen implicitly.

google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
rajivpb: Yep! Fixed, based on what the table name is.

google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
rajivpb: Done for all operators now.

google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
rajivpb: This was some defensive/paranoid coding. Removed now.

google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
rajivpb: Done.

google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
rajivpb: Was looking into this some more and realized that the previous response was incomplete: 

 - If the pipeline is misconfigured, we will indeed catch it at pipeline definition time once the bug on jsonschema validation is fixed: https://github.com/googledatalab/pydatalab/issues/499

 - If there is as failure to have an output table during execute an exception will get thrown downstream, so we should never be here. 
https://github.com/googledatalab/pydatalab/blob/64456d040bb7c856ee99f75bc5a1f179b6dabe0b/google/datalab/bigquery/_query_job.py#L109

Either way, removing the if block and not throwing here makes sense.

google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
```diff
@@ -31,13 +30,15 @@ def __init__(self, sql, parameters=None, table=None, mode=None, *args, **kwargs)
...
     output_options = bq.QueryOutput.table(name=self._table, mode=self._mode, use_cache=False,
-                                          allow_large_results=False)
-    pydatalab_context = google.datalab.Context.default()
+                                          allow_large_results=self._table is not None)
+
     query_params = Pipeline._get_query_parameters(self._parameters)
-    # TODO(rajivpb): Is this a sync or an async operation? Unlike load(), this does not wrap its
-    # async counterpart with a job.wait(). Test and wait if necessary.
-    query.execute(output_options, context=pydatalab_context, query_params=query_params)
+    job = query.execute(output_options, context=None, query_params=query_params)
```
nikhilk: context should have a default; so no need to explicitly pass None.

To be honest, I am not sure why parameter has been added to all these methods like `execute`. I would have expected these to only have been in the ctors for various objects - Query, Table, etc. so as to provide the project/credentials and any other config to use when making API calls pertaining to the object in question.

google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
```diff
@@ -20,26 +20,38 @@ class ExtractOperator(BaseOperator):
...
+      csv_delimiter=self._csv_options.get('delimiter'),
+      csv_header=self._csv_options.get('header'), compress=self._csv_options.get('compress'))
 
     if job.failed:
       raise Exception('Extract failed: %s' % str(job.fatal_error))
     elif job.errors:
       raise Exception('Extract completed with errors: %s' % str(job.errors))
-    return job.result()
+    return {
+      'result': job.result()
```
nikhilk: Maybe the job itself?
As well as the path/pattern of files produced.

google/datalab/contrib/bigquery/operators/_bq_load_operator.py
```diff
@@ -67,3 +65,7 @@ def execute(self, context):
       raise Exception('Load failed: %s' % str(job.fatal_error))
     elif job.errors:
       raise Exception('Load completed with errors: %s' % str(job.errors))
+
+    return {
+      'result': job.result()
```
nikhilk: Maybe the job itself.
And the resulting table in addition.

google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
```diff
@@ -31,13 +30,15 @@ def __init__(self, sql, parameters=None, table=None, mode=None, *args, **kwargs)
...
+
+    # Returning the table-name here makes it available for downstream task instances.
+    return {
+      'table': job.result().name
```
nikhilk: Perhaps the job as well?

google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
rajivpb: Removed context now.

Agreed, I wondered about that too. Perhaps we can sweep it up in refactoring-fixit.

google/datalab/contrib/bigquery/operators/_bq_execute_operator.py
rajivpb: I'm a little hesitant to put non-primitive types here because I'm not sure if airflow's xcom can handle this, so I'm just being safe. It essentially writes all these values to the db, and I'm nervous about any pickling that it might do for this under the covers.

google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
rajivpb: See comment above about job. 

On the path/patterns, why? Won't this be available from the spec itself?

google/datalab/contrib/bigquery/operators/_bq_load_operator.py
rajivpb: See concern above about using xcom for communicating non-primitive types. 

In general, we can go about this in two ways - using xcom to communicate all information, and using the input spec to statically get all the information we need. Agreed that the former gives us more flexibility. However, it seems like something we could do later, once this is more stable. 

Filed P1 https://github.com/googledatalab/pydatalab/issues/569

rajivpb:
https://github.com/googledatalab/pydatalab/pull/577 cloned this and was closed.

rajivpb:
https://github.com/googledatalab/pydatalab/pull/577

PR #567, opened 2017-10-23

rajivpb:
Clone of https://github.com/googledatalab/pydatalab/pull/566

Creating because the above PR has run into CLA issues (I pushed this from my Mac after having forgotten to set the user name and email properly; hopefully this works).


rajivpb:
Was hoping this will pass the CLA checks after I set the user-name and email in my git config, but alas that is not to be. Closing this. Will create another clone of #566 once it's approved.

Issue #546, opened 2017-10-09
Add support for external_table_definition support in Datalab


rajivpb:
https://cloud.google.com/bigquery/external-table-definition

This will enable querying data (like log files) that are stored in GCS, and writing out the results into GCS via the temp-table that BQ creates for this.

rajivpb:
#579

PR #571, opened 2017-10-24
%bq pipeline supports external data sources

rajivpb:
https://github.com/googledatalab/pydatalab/issues/546


rajivpb:
Cloned by #579

PR #575, opened 2017-10-26
Add "%%ml probe" for another way to explain image prediction. It is based on integrated gradients theory.

qimingj:
Also fixed a few py3 errors. Suppress some TF warnings.


google/datalab/contrib/mlworkbench/commands/_ml.py
```diff
@@ -415,6 +417,26 @@ def ml(line, cell=None):
                                    help='Prediction Data. Can be a csv line, or a dict.')
   explain_parser.set_defaults(func=_explain)
 
+  probe_parser = parser.subcommand(
+      'probe',
```
chmeyers: The name "probe" for this is not immediately obvious to me.  Why did you choose it?

PR #577, opened 2017-10-27

rajivpb:
Clone of https://github.com/googledatalab/pydatalab/pull/566 since that ran into CLA issues.

### DIFF
--- a/google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
+++ b/google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
@@ -52,6 +52,3 @@ class ExtractOperator(BaseOperator):
       raise Exception('Extract failed: %s' % str(job.fatal_error))
     elif job.errors:
       raise Exception('Extract completed with errors: %s' % str(job.errors))
-    return {
-      'result': job.result()
-    }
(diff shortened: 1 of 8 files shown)


google/datalab/contrib/bigquery/operators/_bq_extract_operator.py
```diff
@@ -20,26 +20,38 @@ class ExtractOperator(BaseOperator):
...
+      csv_delimiter=self._csv_options.get('delimiter'),
+      csv_header=self._csv_options.get('header'), compress=self._csv_options.get('compress'))
 
     if job.failed:
       raise Exception('Extract failed: %s' % str(job.fatal_error))
     elif job.errors:
       raise Exception('Extract completed with errors: %s' % str(job.errors))
-    return job.result()
+    return {
+      'result': job.result()
```
nikhilk: Is job.result a simple type?
I think extract the list (or something that indicates) the set of files created ... scenario: a subsequent operator wants to act on the resulting files.

google/datalab/contrib/bigquery/operators/_bq_load_operator.py
```diff
@@ -67,3 +65,7 @@ def execute(self, context):
       raise Exception('Load failed: %s' % str(job.fatal_error))
     elif job.errors:
       raise Exception('Load completed with errors: %s' % str(job.errors))
+
+    return {
+      'result': job.result()
```
nikhilk: Same thing here. Is job.result a simple type?
And shouldn't this also return the table into which data was loaded?

PR #585, opened 2017-11-02
Set up script for Composer and support for getting gcsDagLocation from Composer environment 

rajivpb:


google/datalab/contrib/pipeline/composer/setup.sh
```diff
@@ -0,0 +1,25 @@
+ZONE=us-central1
+PROJECT=$1
+EMAIL=$2
+ZONE=$3
+ENVIRONMENT=$4
+
+
+gcloud config set project $PROJECT
+gcloud auth login --activate $EMAIL
+sudo gcloud components repositories add https://storage.googleapis.com/composer-trusted-tester/components-2.json
```
fenglu-g: First thing is to install composer specific gcloud and it may vary slightly depending on the environment (e.g., VM, shell, local) before any gcloud * operations.

google/datalab/contrib/pipeline/composer/setup.sh
```diff
@@ -0,0 +1,25 @@
...
+
+#gcloud components update -q
+sudo apt-get update
+sudo apt-get --only-upgrade install kubectl google-cloud-sdk google-cloud-sdk-datastore-emulator \
+google-cloud-sdk-pubsub-emulator google-cloud-sdk-app-engine-go google-cloud-sdk-app-engine-java \
+google-cloud-sdk-app-engine-python google-cloud-sdk-cbt google-cloud-sdk-bigtable-emulator \
+google-cloud-sdk-datalab
```
fenglu-g: We can put apt-get first, and then composer gcloud installation, and finally composer related configs and environment creation.

google/datalab/contrib/pipeline/airflow/setup.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+PROJECT=${1:-cloud-ml-dev}
+EMAIL=${2:-rajivpb@google.com}
+ZONE=${3:-us-central1}
+ENVIRONMENT=${3:-rajivpb-airflow}
```
qimingj: Should it be 4?

chmeyers: please fix.

google/datalab/contrib/pipeline/composer/_composer.py
```diff
@@ -29,22 +30,34 @@ def __init__(self, zone, environment):
...
-  @property
-  def get_bucket_name(self):
-    # environment_details = Api().environment_details_get(self._zone, self._environment)
-
-    # TODO(rajivpb): Get this programmatically from the Composer API
-    return 'airflow-staging-test36490808-bucket'
+  def gcs_dag_location(self):
+    if not self._gcs_dag_location:
+      environment_details = Api.environment_details_get(self._zone, self._environment)
+      if 'config' not in environment_details \
```
chmeyers: Per style guide, don't use \ for line continuation.  Wrap the statement in parens instead.

google/datalab/contrib/pipeline/composer/_composer.py
```diff
@@ -29,22 +30,34 @@ def __init__(self, zone, environment):
...
-    bucket = client.get_bucket(self.bucket_name)
-    filename = 'dags/{0}.py'.format(name)
+    try:
+      gcs_dag_location_splits = self.gcs_dag_location.split('/')
+      bucket_name = gcs_dag_location_splits[2]
+      # Usually the splits are like ['gs:', '', 'foo_bucket', 'dags']. But we could have additional
+      # parts after the bucket. In those cases, the final file path needs to include those as well
+      additional_parts = ''
+      if len(gcs_dag_location_splits) > 4:
+        additional_parts = '/' + '/'.join(gcs_dag_location_splits[4:])
```
chmeyers: Instead of reconstructing this, just use the maxsplit argument to split so that the path section won't be split in the first place.

https://docs.python.org/2/library/stdtypes.html#str.split

google/datalab/contrib/pipeline/composer/_composer.py
```diff
@@ -29,22 +30,34 @@ def __init__(self, zone, environment):
...
     self._zone = zone
     self._environment = environment
+    self._gcs_dag_location = None
 
   def deploy(self, name, dag_string):
     client = gcs.Client()
-    bucket = client.get_bucket(self.bucket_name)
-    filename = 'dags/{0}.py'.format(name)
+    try:
+      gcs_dag_location_splits = self.gcs_dag_location.split('/')
```
chmeyers: _ , _, bucket, filepath = self.gcs_dag_location.split('/', 3)

You should have gcs_dag_location() verify that what it returns is in the expected format with trailing slash, etc.  And test whether this works with a top level directory in the bucket (i.e. gcs_dag_location = "gs://bucket/" )

google/datalab/contrib/pipeline/composer/setup.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Recycling shell to pick up the new gcloud.
+exec -l $SHELL
+
+# These have to be set here and not on the top of the script because we recycle the shell somewhere
+# between the start of this script and here.
+PROJECT=${1:-cloud-ml-dev}
+EMAIL=${2:-rajivpb@google.com}
+ZONE=${3:-us-central1}
+ENVIRONMENT=${3:-datalab-testing-1}
```
chmeyers: 4 instead of 3 here?

google/datalab/contrib/pipeline/composer/setup.sh
```diff
@@ -0,0 +1,37 @@
...
+
+gcloud config set project $PROJECT
+gcloud config set account $EMAIL
+gcloud auth login --activate $EMAIL
```
chmeyers: --activate is the default

Is this setup.sh only ever run manually?  (Asking because users shouldn't gcloud auth login on a VM.)

release.sh
```diff
@@ -16,6 +16,7 @@
 # Compiles the typescript sources to javascript and submits the files
 # to the pypi server specified as first parameter, defaults to testpypi
 # In order to run this script locally, make sure you have the following:
+# - A Python 3 environment (due to urllib issues)
```
chmeyers: Is there an issue open to fix this for Py2?

google/datalab/contrib/pipeline/composer/_composer.py
```diff
@@ -29,22 +30,34 @@ def __init__(self, zone, environment):
...
-    # TODO(rajivpb): Get this programmatically from the Composer API
-    return 'airflow-staging-test36490808-bucket'
+  def gcs_dag_location(self):
+    if not self._gcs_dag_location:
+      environment_details = Api.environment_details_get(self._zone, self._environment)
+      if 'config' not in environment_details \
+              or 'gcsDagLocation' not in environment_details.get('config'):
+        raise ValueError('Dag location unavailable from Composer environment {0}'.format(
+          self._environment))
+      self._gcs_dag_location = environment_details['config']['gcsDagLocation']
```
chmeyers: Add a slash to the end of the location if it doesn't end with a slash?

google/datalab/contrib/pipeline/airflow/setup.sh
rajivpb: Correct; fixed now.

release.sh
rajivpb: https://github.com/googledatalab/pydatalab/issues/605

google/datalab/contrib/pipeline/composer/_composer.py
rajivpb: Fixed.

google/datalab/contrib/pipeline/composer/_composer.py
rajivpb: Ah, cool. Done.

google/datalab/contrib/pipeline/composer/_composer.py
rajivpb: Done.

google/datalab/contrib/pipeline/composer/_composer.py
rajivpb: That made my life easy, thanks!

google/datalab/contrib/pipeline/composer/setup.sh
rajivpb: Done.

google/datalab/contrib/pipeline/composer/setup.sh
rajivpb: Yes, setup.sh is only ever run manually on a local machine at the moment. It's not for the VM scenario.

rajivpb: (Removed --activate)

google/datalab/contrib/pipeline/composer/_composer.py
```diff
@@ -34,30 +37,29 @@ def __init__(self, zone, environment):
...
           self._environment))
-      self._gcs_dag_location = environment_details['config']['gcsDagLocation']
+      gcs_dag_location = environment_details['config']['gcsDagLocation']
+
+      if gcs_dag_location is None or not self.gcs_file_regexp.match(gcs_dag_location):
+        raise ValueError(
+          'Dag location {0} from Composer environment {1} is in incorrect format'.format(
+            gcs_dag_location, self._environment))
+
+      self._gcs_dag_location = gcs_dag_location + '/'
```
chmeyers: What if it already ends with a slash?

rajivpb: Initially disallowed it via regex, but later decided to be more tolerant. Inserted check now.

google/datalab/contrib/pipeline/airflow/setup.sh
qimingj: I think you fixed the composer one but not airflow one yet?

google/datalab/contrib/bigquery/commands/_bigquery.py
```diff
@@ -21,6 +20,10 @@ def _create_pipeline_subparser(parser):
                                                   'transform data using BigQuery.')
   pipeline_parser.add_argument('-n', '--name', type=str, help='BigQuery pipeline name',
                                required=True)
+  pipeline_parser.add_argument('-e', '--environment', type=str,
+                               help='The name of the Composer or Airflow environment.')
```
qimingj: It's only the name of the Composer right? Not used in Airflow?

google/datalab/contrib/bigquery/commands/_bigquery.py
```diff
@@ -44,11 +47,17 @@ def _pipeline_cell(args, cell_body):
...
         cell_body, utils.commands.notebook_environment())
     pipeline_spec = _get_pipeline_spec_from_config(bq_pipeline_config)
+    # TODO(rajivpb): This import is a stop-gap for
+    # https://github.com/googledatalab/pydatalab/issues/593
+    import google.datalab.contrib.pipeline._pipeline
     pipeline = google.datalab.contrib.pipeline._pipeline.Pipeline(name, pipeline_spec)
     utils.commands.notebook_environment()[name] = pipeline
 
     # If a composer environment and zone are specified, we deploy to composer
     if 'environment' in args and 'zone' in args:
```
qimingj: I think "'environment' in args and 'zone' in args" is always True. If user does not provide values for them, they will be None but still 'environment' and 'zone' exist in args.

Perhaps you want  "if args['environment'] and args[ 'zone']:".

google/datalab/contrib/bigquery/commands/_bigquery.py
rajivpb: You're right. At the of putting the comment, I made it slightly future looking since we plan to work on an Airflow environment. Will change in a different PR mostly because I'm trying to get this in before demo time today.

google/datalab/contrib/bigquery/commands/_bigquery.py
rajivpb: Oh, I didn't know that. Will verify and fix in a different PR mostly because I'm trying to get this in before demo time today.

google/datalab/contrib/pipeline/airflow/setup.sh
rajivpb: Yeah, my bad - I kept thinking that all the comments referred to the composer's setup.sh and was a little confused on getting multiple comments for the same change. Will fix in a different PR mostly because I'm trying to get this in before demo time today.

Issue #597, opened 2017-11-04
Replace all newlines and tabs in a SQL query with spaces.


rajivpb:

rajivpb:
https://github.com/googledatalab/pydatalab/pull/600

PR #600, opened 2017-11-04
Triple quote all string parameters in airflow python spec 

rajivpb:
Some string parameter values can have special characters that need to be escaped when they make their way to the python spec. Triple quoting these values would prevent unexpected behaviour. First saw this when composing sql queries that spanned multiple lines in Datalab.

### DIFF
--- a/google/datalab/contrib/bigquery/commands/_bigquery.py
+++ b/google/datalab/contrib/bigquery/commands/_bigquery.py
@@ -183,9 +183,10 @@ def _get_execute_parameters(load_task_id, bq_pipeline_input_config,
         if 'csv' in bq_pipeline_input_config:
           execute_task_config['csv_options'] = bq_pipeline_input_config.get('csv')
 
-    # Stuff from the transformation config
     query = utils.commands.get_notebook_item(bq_pipeline_transformation_config['query'])
-    execute_task_config['sql'] = query.sql
+    # The user's sql could span multiple lines. Here, we replace new-lines (and the carriage-return)
+    # with spaces because it could result in malformed python.
+    execute_task_config['sql'] = query.sql.replace('\n', ' ').replace('\r', ' ')
 
     # Stuff from the output config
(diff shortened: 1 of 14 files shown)


rajivpb:
#597

google/datalab/contrib/bigquery/commands/_bigquery.py
```diff
@@ -183,9 +183,10 @@ def _get_execute_parameters(load_task_id, bq_pipeline_input_config,
         if 'csv' in bq_pipeline_input_config:
           execute_task_config['csv_options'] = bq_pipeline_input_config.get('csv')
 
-    # Stuff from the transformation config
     query = utils.commands.get_notebook_item(bq_pipeline_transformation_config['query'])
-    execute_task_config['sql'] = query.sql
+    # The user's sql could span multiple lines. Here, we replace new-lines (and the carriage-return)
+    # with spaces because it could result in malformed python.
+    execute_task_config['sql'] = query.sql.replace('\n', ' ').replace('\r', ' ')
```
qimingj: Can we just wrap user's sql with triple quotes in the generated code instead?

rajivpb: Yes, we could. The fix for that will require some thought though. Currently, this gets passed to the underlying object model (i.e. the - soon to be renamed - Pipeline class). There, it gets formatted as a string:
https://github.com/googledatalab/pydatalab/blob/51c668fe8e7658f64e1cfdbc76ebc66c9f6cfbfa/google/datalab/contrib/pipeline/_pipeline.py#L171

I see two choices - we could either have every string parameter be formatted with three quotes. My concern about that it could make the pipeline spec hard to read. The other choice of course is to make the Pipeline object model aware that this is a sql query string, and that it would need to be triple quoted. That breaks encapsulation somewhat. 

This fix so far feels like a better compromise.

PR #614, opened 2017-11-15
Recursively string-format all nested values of the dict in yaml input

rajivpb:
There are cases where the old style format modifiers are actually used in sql for specifying formats (Google SQL has a concept of date-time format strings that use format modifiers with %). This interferes with the resolving of the airflow macros. Moving to the new style (i.e. the curly braces) mitigates this to a large extent. The risk of {} being used in the yaml cell-body remains, but is lower compared to usage of %.


google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -142,14 +142,24 @@ def _get_operator_definition(self, task_id, task_details):
...
 
     return '{0} = {1}({2}, dag=dag)\n'.format(task_id, operator_classname, full_param_string)
 
+  def _resolve_airflow_macros(self, operator_param_value):
+    if type(operator_param_value) is list:
+      return [self._resolve_airflow_macros(item) for item in operator_param_value]
+    if type(operator_param_value) is dict:
+      return dict((self._resolve_airflow_macros(k), self._resolve_airflow_macros(v))
+                  for k, v in operator_param_value.items())
+    if isinstance(operator_param_value, basestring):
```
jimmc: All the other files in pydatalab that use basestring include this line:
from past.builtins import basestring

rajivpb: Ah, thanks! Looks like while it's not a requirement for py2, it's indeed a requirement for py3.

qimingj: Not sure "isinstance(operator_param_value, basestring)" covers all string types in py3.

I would suggest "isinstance(operator_param_value, six.string_types)", which seems popular. Within pydatalab, see https://github.com/googledatalab/pydatalab/search?utf8=%E2%9C%93&q=six.string_types&type=.

google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -142,14 +143,24 @@ def _get_operator_definition(self, task_id, task_details):
...
       param_format_string = Pipeline._get_param_format_string(
           operator_param_value)
       param_string = param_format_string.format(operator_param_name, operator_param_value)
-      param_string = param_string % self._airflow_macros
       full_param_string = full_param_string + param_string
 
     return '{0} = {1}({2}, dag=dag)\n'.format(task_id, operator_classname, full_param_string)
 
+  def _resolve_airflow_macros(self, operator_param_value):
+    if type(operator_param_value) is list:
```
qimingj: isinstance(operator_param_value, list) as is mostly done in pydatalab.

google/datalab/contrib/pipeline/_pipeline.py
```diff
@@ -142,14 +143,24 @@ def _get_operator_definition(self, task_id, task_details):
...
+      return [self._resolve_airflow_macros(item) for item in operator_param_value]
+    if type(operator_param_value) is dict:
+      return dict((self._resolve_airflow_macros(k), self._resolve_airflow_macros(v))
```
qimingj: Why not just:

return {self._resolve_airflow_macros(k): self._resolve_airflow_macros(v) for for k, v in operator_param_value.items()}

to avoid the cast to dict?

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: Done.

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: Great suggestion; done.

google/datalab/contrib/pipeline/_pipeline.py
rajivpb: Done.

PR #615, opened 2017-11-15
Fix a performance issue that it loads require.js in each cell output unnecessarily.

qimingj:
I believe the glocal require js config for each cell was added because we now use an iframe for each output cell in experimental UI.

The issue is worth fixing because:

1. require js appends a unique query string parameter to require.js, which makes the file non-cachable, so each output cell needs to download the file (84KB). If the notebook includes 20+ cells (which is very common), it takes long time to load (15s+ difference on my desktop)

2. Each cell output in the notebook includes a require js config section, which makes the notebook size unnecessarily larger.

The fix is to remove the global require js config for each cell output, and add it to only the cell output that need it.


datalab/bigquery/commands/_bigquery.py
```diff
@@ -910,10 +910,12 @@ def _table_viewer(table, rows_per_page=25, fields=None):
   _HTML_TEMPLATE = u"""
     <div class="bqtv" id="{div_id}">{static_table}</div>
     <br />{meta_data}<br />
+    <script src="/static/components/requirejs/require.js"></script>
```
jimmc: Instead of mixing this stuff into the existing html for every place it is needed, could you do something like the way it was done before using configure_global_state? In other words, keep the function configure_global_state (probably rename it), but instead of registering it as a pre_run_cell hook, call it directly from the places where you have modified the html, and revert the html to what it was. I don't know if that would work.
Or, if you do have to bundle it in with the html in each of these functions, would it work to leave the existing requirejs.config call alone, and add another one that just sets the /static/base path? If so, then you could have a single string that contains the requirejs stuff, and insert it into each html with something like {require_js_script}, where that variable resolves to the string that is currently in configure_global_state.

qimingj: We could. There are two reasons I think it is better to just include it inline:

1. it needs to include a base path config anyway like you mentioned:

      require.config({
          paths: {
            base: '/static/base',
          }
      });

Instead of creating another require.config, it is easier to just mix it with the require config section of each individual output, such as:

require.config({
          paths: {
              base: '/static/base',
              d3: '//cdnjs.cloudflare.com/ajax/libs/d3/3.4.13/d3',
              plotly: 'https://cdn.plot.ly/plotly-1.5.1.min.js?noext',
              ...

Note that existing individual requirejs config is not self contained. It sometimes references the "base" path.

2. I'd like to have all output cell that requires requireJS self-contained. It's a bit hard to maintain to have the HTML split and included in different files, especially there are only 3 places that require requireJS.

Thoughts?

jimmc: I see six places in four files where you are adding the requirejs script. I guess my question is, what is the likelihood that there will be any changes to that stuff, such as a change to /static/base or the path to requirejs, which would require finding each place that has that code fragment and updating it?
I'm not particularly familiar with the requirejs stuff, so I don't have a strong argument to make on how it should be done. If you think this way is best, I'm OK with that.

qimingj: It is six. Sorry. And six because we duplicated the legacy datalab lib. Functionally it is 3 (that's why I had 3).

I agree that It is a trade-off between being self-contained and code reusability. I don't feel strongly that the following as an independent code unit:

<script src="/static/components/requirejs/require.js"></script>
<script>
     requirejs.config({
              paths: {
                base: '/static/base',
              },
            });
</script>

It does not save much source code, because these output has a "require.config" section anyway. If we ever want to change anything from the above, chances are that we may need to change the paths in each output cell that require RequireJS anyway.

PR #617, opened 2017-11-16
Delete composer and airflow scripts (Part 1 of move to googledatalab/notebooks)

rajivpb:
The create script will be incorporated into a Datalab notebook, and the startup-script will be uploaded to a well-known GCS location (temporary ones being used for now).


google/datalab/contrib/pipeline/airflow/airflow_vm_create.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+VM_NAME="instance-7"
```
nikhilk: These should be options/flags.

google/datalab/contrib/pipeline/airflow/airflow_vm_create.sh
```diff
@@ -0,0 +1,30 @@
...
+PROJECT_ID="cloud-ml-dev"
+ZONE="us-central1-b"
+
+# TODO(rajivpb): These should have more official values
+AIRFLOW_VM_STARTUP_SCRIPT="gs://rajivpb-airflow-testing/airflow_vm_startup.sh"
```
nikhilk: The bucket should also be a flag.

google/datalab/contrib/pipeline/airflow/airflow_vm_create.sh
```diff
@@ -0,0 +1,30 @@
...
+GCS_DAG_BUCKET_FOR_AIRFLOW="rajivpb-airflow-testing"
+GCS_DAG_PATH_FOR_AIRFLOW="/dags"
+
+# The default values here were grabbed from Pantheon's "command-line" button for creating a VM
+gcloud beta compute --project $PROJECT_ID instances create $VM_NAME \
+--zone $ZONE \
```
nikhilk: Nit - Slightly more readable if these were indented IMO.

google/datalab/contrib/pipeline/airflow/airflow_vm_create.sh
```diff
@@ -0,0 +1,30 @@
...
+--machine-type "n1-standard-1" \
+--network "default" \
+--maintenance-policy "MIGRATE" \
+--service-account "236417448818-compute@developer.gserviceaccount.com" \
```
nikhilk: Should just use the default service account. Or it would need to be a flag.

google/datalab/contrib/pipeline/airflow/airflow_vm_create.sh
```diff
@@ -0,0 +1,30 @@
...
+--image-project "debian-cloud" \
+--boot-disk-size "10" \
+--boot-disk-type "pd-standard" \
+--boot-disk-device-name $VM_NAME \
+--metadata-from-file startup-script-url=$AIRFLOW_VM_STARTUP_SCRIPT \
+
+# TODO(rajivpb): Eventually, we need to set this metadata so that they can be parameters
+# --metadata GCS_DAG_BUCKET=$GCS_DAG_BUCKET_FOR_AIRFLOW,GCS_DAG_PATH=$GCS_DAG_PATH_FOR_AIRFLOW \
+
+gcloud compute --project $PROJECT_ID ssh --zone $ZONE $VM_NAME
```
nikhilk: Is there a reason why the script starts an SSH session?

google/datalab/contrib/pipeline/airflow/airflow_vm_startup.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+LOCAL_DAG_PATH=airflow/dags
```
nikhilk: You should probably make these explicitly root relative. So /airflow/dags (leading slash).

google/datalab/contrib/pipeline/airflow/airflow_vm_startup.sh
```diff
@@ -0,0 +1,27 @@
...
+# been set by the VM's create script. So eventually, remove these lines.
+GCS_DAG_BUCKET="rajivpb-airflow-testing"
+GCS_DAG_PATH="/dags"
+
+export AIRFLOW_HOME=airflow
+
+apt-get --assume-yes install python-pip
+
+pip install airflow
+.local/bin/airflow initdb
```
nikhilk: Unclear why you need the .local/bin. Doesn't pip install put it in the path? Or should you change $PATH to make sure airflow is on it?

google/datalab/contrib/pipeline/airflow/airflow_vm_startup.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+LOCAL_DAG_PATH=airflow/dags
+TEMP_CRON_FILE=temp_crontab.txt
```
nikhilk: Nit -- but, why TEMP? Just call it AIRFLOW_CRON or something...

google/datalab/contrib/pipeline/airflow/airflow_vm_startup.sh
```diff
@@ -0,0 +1,27 @@
...
+gsutil cp gs://rajivpb-airflow-testing/datalab-1.1.0.tar .
+pip install datalab-1.1.0.tar
+rm datalab-1.1.0.tar
+
+# We append a gsutil rsync command to the cron file and have this run every minute to sync dags.
+mkdir -p $LOCAL_DAG_PATH
+crontab -l > $TEMP_CRON_FILE
+echo "* * * * * /usr/bin/gsutil rsync gs://$GCS_DAG_BUCKET/$GCS_DAG_PATH $LOCAL_DAG_PATH" >> $TEMP_CRON_FILE
+crontab $TEMP_CRON_FILE
+rm $TEMP_CRON_FILE
```
nikhilk: Don't delete -- let users see/modify etc.

google/datalab/contrib/pipeline/airflow/airflow_vm_create.sh
rajivpb: Well, I'm actually writing this script to be a set of commands for some easy testing which I can easily copy-paste into a notebook. Hence I've not used flags. It's not actually meant for being executed and I will delete it once the notebook is ready.

google/datalab/contrib/pipeline/airflow/airflow_vm_create.sh
rajivpb: See comment above.

google/datalab/contrib/pipeline/airflow/airflow_vm_create.sh
nikhilk: We can just give alpha users a set of .sh files. So that way, what you use for testing is what we give out.

google/datalab/contrib/pipeline/airflow/airflow_vm_create.sh
rajivpb: I've left it there for convenience for testing.

google/datalab/contrib/pipeline/airflow/airflow_vm_create.sh
rajivpb: This PR is not for submission and is only for review since this isn't fully working yet. Updated title accordingly.

google/datalab/contrib/pipeline/airflow/airflow_vm_create.sh
rajivpb: Yep. Removed now.

google/datalab/contrib/pipeline/airflow/airflow_vm_startup.sh
rajivpb: Curiously enough, pip install did not put it in the path. So I saw a need to do this. Didn't want to add it to the path and make changes to the environment variables since I was using this in only one place, so did it this way.

google/datalab/contrib/pipeline/airflow/airflow_vm_startup.sh
rajivpb: But users can always do crontab -l to view, right?

google/datalab/contrib/pipeline/airflow/airflow_vm_startup.sh
nikhilk: See https://airflow.readthedocs.io/en/1.6.2-docs/start.html ... about AIRFLOW_HOME, and apparent lack of needing to use .local/bin or updating $PATH.

google/datalab/contrib/pipeline/airflow/airflow_vm_create.sh
rajivpb: Hmm ... possibly. I have a slight preference for giving out a notebook though.

PR #619, opened 2017-11-21
Add macros for year, month, day, etc. and merge these with airflow macros

rajivpb:


rajivpb:
This, as it turns out, does not work as expected for SQL parameters (the '{{ execution_date.xxx }}' where xxx = year/month/day does not work when it's a param value in the query_params dict). I am looking into alternate implementations and there are a few promising leads. I'm merging this in since it does work for the GCS paths. I'll tackle the SQL params issue with a separate PR.

PR #624, opened 2017-11-28
Use a few canned query parameters for BigQuery

rajivpb:
Not unit-tested/verified, not-yet-prettified, etc. Sending PR because I want to make sure that the general approach is agreeable (in terms of contrib-related code in core, etc.).


rajivpb:
Not unit-tested/verified, not-yet-prettified, etc. Sending PR because I want to make sure that the general approach is agreeable (in terms of contrib-related code in core, etc.).

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -376,17 +376,45 @@ def _get_query_parameters(args, cell_body):
   if config:
     jsonschema.validate(config, BigQuerySchema.QUERY_PARAMS_SCHEMA)
 
+    # We merge the parameters with the airflow macros so that users can specify certain airflow
+    # macro names (like '@ds') in their sql. This is useful for enabling the user to progressively
+    # author a pipeline with %bq pipeline.
+
+    import datetime
+    canned_parameters = {
```
nikhilk: You probably want to get the timestamp value once, and format all versions using that value.

nikhilk: As I look at this, I wonder if these should be prefixed in some way to avoid potential conflicts with user-defined parameters having similar name. Example... `ds` and `ts` don't seem like the ones potentially already in use.

nikhilk: pipeline_parameters would be a more descriptive name.

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -376,17 +376,45 @@ def _get_query_parameters(args, cell_body):
...
+      'ts_hour': {'name': 'ds', 'type': 'STRING', 'value': datetime.datetime.now().strftime('%H')},
+      'ts_minute': {'name': 'ds', 'type': 'STRING', 'value': datetime.datetime.now().strftime('%M')},
+      'ts_second': {'name': 'ds', 'type': 'STRING', 'value': datetime.datetime.now().strftime('%S')},
+    }
+
+    # We merge the parameters by first pushing in the query parameters into a dictionary keyed by
+    # the parameter name. We then use this to update the canned parameters dictionary. This will
+    # have the effect of using user-provided parameters in case of naming conflicts.
+    parameters = {item['name']: item for item in config['parameters']}
+    merged_parameters = canned_parameters.update(parameters)
```
nikhilk: Have you validated that BQ is fine with extra unused parameters to ensure this all holds up.

google/datalab/bigquery/commands/_bigquery.py
```diff
@@ -376,22 +377,54 @@ def _get_query_parameters(args, cell_body):
...
-          'value': param['value']
-        }
-      })
-    return parsed_params
-  else:
-    return {}
+  # We merge the parameters with the airflow macros so that users can specify certain airflow
+  # macro names (like '@ds') in their sql. This is useful for enabling the user to progressively
+  # author a pipeline with %bq pipeline.
+  today = datetime.date.today()
```
nikhilk: This would be slightly better, so you have just one timestamp that everything is based off of:

    today = now.date()

PR #625, opened 2017-11-29
Making it possible to use SQL udfs

freyrsae:
The syntax for SQL UDFs varies from the js one and currently it is not possible to use the UDF class to create SQL UDFs. This PR addresses this problem.

For example, doing:

`
addOne = bq.UDF(
    name="addOne",
    code="i + 1",
    return_type="INT64",
    language="SQL",
    params=[("i", "INT64")]
)

test_query = bq.Query('''
SELECT
  addOne(i)
FROM
  (SELECT 1 as i)

''', udfs={"addOne": addOne})
`
is currently parsed to 

`
CREATE TEMPORARY FUNCTION addOne (i INT64)
RETURNS INT64
LANGUAGE SQL
AS """
i + 1
"""
OPTIONS (

);

SELECT
  addOne(i)
FROM
  (SELECT 1 as i)
`

which does not work and should be 
`
CREATE TEMPORARY FUNCTION addOne (i INT64)
RETURNS INT64
AS (
i + 1
);

SELECT
  addOne(i)
FROM
  (SELECT 1 as i)
`


google/datalab/bigquery/_udf.py
```diff
@@ -96,14 +96,21 @@ def _build_udf(name, code, return_type, params, language, imports):
...
+    if language.lower() == 'sql':
+        udf = 'CREATE TEMPORARY FUNCTION {name} ({params})\n' + \
+              'RETURNS {return_type}\n' + \
+              'AS (\n' + \
+              '{code}\n' + \
+              ');'
+    else:
+        udf = 'CREATE TEMPORARY FUNCTION {name} ({params})\n' +\
+              'RETURNS {return_type}\n' +\
+              ('LANGUAGE {language}\n' if language.lower() != 'sql' else '') +\
```
nikhilk: Presumably the language is always != 'sql' at this point, and can be simplified. I doubt the generated LANGUAGE \n string is going to work in fact.

freyrsae: Right you are, have added a fix

freyrsae:
@nikhilk I've fixed the requested changes, anything else?

freyrsae:
@nikhilk or anyone (@qimingj perhaps), any more blockers on merging this?

PR #626, opened 2017-11-30
Add xgboost package to ml workbench.

qimingj:
Only preprocessing is added. training/prediction will come later. Analysis and transform steps are mostly inheritted from tensorflow ones, with the exception that it generates libsvm file as transformed results, which can be loaded by xgb directly.

Also rename tests/mltoolbox_code_free_ml to tests/ml_workbench.

### DIFF
--- a/solutionbox/ml_workbench/xgboost/setup.py
+++ b/solutionbox/ml_workbench/xgboost/setup.py
@@ -28,7 +28,7 @@ setup(
   """,
   install_requires=[
     'tensorflow==1.2.0',
-    'protobuf==3.1.0',
+    'protobuf==3.4.0',
     'pillow==3.4.1',  # ML Engine does not have PIL installed
     'xgboost==0.6a2',
   ],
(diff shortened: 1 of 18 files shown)


solutionbox/ml_workbench/xgboost/setup.py
```diff
@@ -0,0 +1,38 @@
...
+  author='Google',
+  author_email='google-cloud-datalab-feedback@googlegroups.com',
+  keywords=[
+  ],
+  license="Apache Software License",
+  long_description="""
+  """,
+  install_requires=[
+    'tensorflow==1.2.0',
+    'protobuf==3.1.0',
```
chmeyers: We can't use a newer version of protobuf?

solutionbox/ml_workbench/xgboost/trainer/feature_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+            'source_column': schema_name}
+      elif schema_type == constant.STRING_SCHEMA:
+        features[schema_name] = {
+            'transform': constant.DEFAULT_CATEGORICAL_TRANSFORM,
+            'source_column': schema_name}
+      else:
+        raise NotImplementedError('Unknown type %s' % schema_type)
+
+
+# TODO(brandondutra): introduce the notion an analysis plan/classes if we
```
chmeyers: Make this TODO for someone other than brandon

PR #627, opened 2017-11-30
Introduce 'load' and 'extract' as aliases for 'input' and 'output'

rajivpb:


google/datalab/contrib/bigquery/commands/_bigquery.py
```diff
@@ -97,9 +92,9 @@ def _get_pipeline_spec_from_config(bq_pipeline_config):
   if email_config:
     pipeline_spec['emails'] = email_config
 
-  input_config = bq_pipeline_config.get('input')
+  input_config = bq_pipeline_config.get('input') or bq_pipeline_config.get('load')
   transformation_config = bq_pipeline_config.get('transformation')
-  output_config = bq_pipeline_config.get('output')
+  output_config = bq_pipeline_config.get('output') or bq_pipeline_config.get('export')
```
nikhilk: Is it %bq export, or %bq extract? I thought it was latter. We should be consistent.

rajivpb: Agreed. I was surprised that you wanted export (instead of extract) in your email. Changed now.

nikhilk: OK, my bad then in email.

PR #630, opened 2017-12-01
Make schedule-related pipeline defaults configurable

rajivpb:


google/datalab/contrib/bigquery/commands/_bigquery.py
```diff
@@ -29,7 +29,9 @@ def _create_pipeline_subparser(parser):
   pipeline_parser.add_argument('-e', '--environment', type=str,
                                help='The name of the Google Cloud Composer environment.')
   pipeline_parser.add_argument('-l', '--location', type=str,
-                               help='The location of the Google Cloud Composer environment.')
+                               help='The location of the Google Cloud Composer environment. '
+                                    'Refer https://cloud.google.com/about/locations/ for further'
```
qimingj: Need an extra space after 'further'.

rajivpb: Good catch, and done. Thanks and PTAL.

PR #637, opened 2017-12-14
Introduce user-defined parameters (and airflow macros) for path and table args across %bq's sub-commands

rajivpb:


chmeyers:
There don't seem to be any Docstring changes showing how these parameters are supposed to work?

PR #638, opened 2017-12-16
Reference table in the input section within SQL via 'input' 

rajivpb:


google/datalab/bigquery/_query.py
```diff
@@ -368,7 +368,8 @@ def get_query_parameters(config_parameters, date_time=datetime.datetime.now()):
     return parsed_params
 
   @staticmethod
-  def resolve_parameters(value, parameters, date_time=datetime.datetime.now()):
+
+  def resolve_parameters(value, parameters, date_time=datetime.datetime.now(), macros=False):
```
qimingj: Add "macros" to docstring.

google/datalab/bigquery/_table.py
```diff
@@ -458,6 +459,10 @@ def load_async(self, source, mode='create', source_format='csv', csv_options=Non
     Raises:
       Exception if the load job failed to be started or invalid arguments were supplied.
     """
+
+    # Resolve any macros necessary
+    source = source
```
qimingj: This should be removed?

google/datalab/bigquery/_query.py
rajivpb: Done.

google/datalab/bigquery/_table.py
rajivpb: Done.

PR #650, opened 2018-01-09
Disable ML Workbench Cloud tests (for now).

qimingj:


rajivpb:
(BTW, this was super helpful in unblocking my work - thanks a bunch!)

Issue #628, opened 2017-12-01
Provide option to shut off backfill


rajivpb:
This is 'on' by default: https://airflow.apache.org/scheduler.html#backfill-and-catchup

rajivpb:
@nikhilk @Di-Ku @chmeyers : The way to do this would be disallow start_dates that are in the past, and to force-overwrite those that are indeed in the past with datetime.now(), i.e. the time at which the notebook cell is executed. Airflow creates dag-runs in the past if start_date + (N*interval) imply that there could have been dag-runs in the past. However, if the dag were re-deployed (with different task-ids and/or different tasks) with the same dag-id, start-date and interval, the dag-runs are _not_ re-run. Essentially, this makes it possible to land up in a state with the same dag-id and 

Does that make sense? 

Note that there is a separate "backfill" option in the airflow CLI, which takes a start_date and an end_date, and creates task-instances for all the intervals in-between. 

In general, I don't think that this should be a ship-blocker, so I'm making this a P1. Let me know if you disagree.

rajivpb:
Turns out that this can be controlled with catchup=False by default. 
https://github.com/googledatalab/pydatalab/pull/651

PR #651, opened 2018-01-09
Add catchup option to prevent backfilling by default

rajivpb:


rajivpb:
https://github.com/googledatalab/pydatalab/issues/628

PR #656, opened 2018-01-11
Changing default of catchup from False to True

rajivpb:
This is so that users can completely leave out the 'schedule' section and expect that their pipeline will run once. 

Currently this is broken because:
1. When the user schedules a pipeline, the start-date is assumed to be now, the end-date is assumed to be infinity, and interval is assumed to be 'once'.
2. Typically, a few minutes have passed between the time the user executes the bq-pipeline cell in Datalab and the time the airflow scheduler picks it up.
3. Because catchup=False, airflow will not execute the pipeline because the the start-date is a few minutes in the past. 

To enable the convenience feature of a user scheduling a pipeline  to 'run-once' without specifying a schedule section at all, we need to the default value for catchup to be True. When the user does not want this sort of "backfill", they will need to specify catchup as False. This is just as well since the user would be specifying a schedule section with defined start-date and end-date values anyway, so adding a 'catch-up' flag would not be much of an inconvenience.

### DIFF
--- a/google/datalab/contrib/bigquery/commands/_bigquery.py
+++ b/google/datalab/contrib/bigquery/commands/_bigquery.py
@@ -53,6 +53,7 @@ Creates a GCS/BigQuery ETL pipeline. The cell-body is specified as follows:
     start: <formatted as '%Y-%m-%dT%H:%M:%S'; default is 'now'>
     end:  <formatted as '%Y-%m-%dT%H:%M:%S'; default is 'forever'>
     interval: {@once (default) | @hourly | @daily | @weekly | @ monthly | @yearly | <cron ex>}
+    catchup: {True (default) | False}; Whether to have pipelines execute in the past.
   parameters: <syntax is same as that in '%%bq execute'>')
 """)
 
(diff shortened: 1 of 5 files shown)


nikhilk:
Default behavior should not inadvertently cause a bunch of work to get queued. Entirely possible that some of the queries could be expensive. Secondly the nature of notebooks is you have some start date in the notebook, and re-running the cell at some point in the future without changing the cell would definitely trigger bunch of work.

I suggest we figure out what to do to make schedule-less spec result in reliable one-time execution. For example start_date could be based on datetime.now() (assuming yesterday as start_date doesn't work, though I thought it did).

nikhilk:
It may be a good idea to have schedule be required -- it leaves too much unclear when its not explicit, esp. if you are opening/reading an existing notebook.

rajivpb:
Yeah, +1 to having schedule be required. I think it would make things simpler for the user (importantly) and also internally in terms of the implementation. And having catchup=False (which is now the default) would prevent the pipelines getting scheduled unnecessarily (at a great expense to the user, potentially, as you mention). 

I'll close this PR without submitting.

PR #684, opened 2018-05-14
Error fixed when installing on python 3.

piffall:


piffall:
I signed it!

ojarjur:
@piffall Can you give more context on what error this fixes?

piffall:
Sure @ojarjur . It was failing when I open the PR, but not now ¬¬. Python 3 should not install this package. 

The error was triggered by this if statement.
https://github.com/agronholm/pythonfutures/blob/master/setup.py#L8

I can't reproduce the error, but I'm not sure why.

ojarjur:
@piffall It looks like the error is dependent on the version of futures that is installed.

If you already have a version prior to 3.2 installed, then this will just be a warning, but if you don't then it will be an error.

If you run `pip freeze` do you see a version of futures installed from before 3.2?

Regardless, given that this will produce an error for some users, I think your change is the right thing to do.

ojarjur:
Thanks for the fix!

piffall:
Thanks!

Issue #686, opened 2018-05-28
google.datalab.utils.Iterator.__iter__ stuck in "infinite" loop


j-planet:
## Cause
The `count` argument of `_get_row_fetcher` should be the running count: https://github.com/googledatalab/pydatalab/blob/ec06b2c1fa4b41be75c7af18c07562afc0a609c3/google/datalab/bigquery/_table.py#L533-L536

But `google.datalab.utils.Iterator.__iter__` is passing in a constant count instead: https://github.com/googledatalab/pydatalab/blob/eec5048c16f1ea0d7be9c2ac5ec49297be6995ae/google/datalab/utils/_iterator.py#L36

## Fix
Change https://github.com/googledatalab/pydatalab/blob/eec5048c16f1ea0d7be9c2ac5ec49297be6995ae/google/datalab/utils/_iterator.py#L40-L41 to `self._count += len(items)`

## How to reproduce the error
The following code is *supposed* to retrieve only 1025 rows. Instead, it first retrieves 1024 rows an HTTP call, then 1 row per HTTP call until the end of the table.

```python
import google.datalab.bigquery as bq
table = bq.Table('some table with > 2000 rows')

for i, row in enumerate(table.range(start_row=1, max_rows=1025)):
    print(i, row)
```

j-planet:
Spent a glorious Sunday night figuring this out. It'd be great if Google can for the Datalab instance costs occurred during those debugging as well as costs caused by an overnight job that ran indefinitely. :)

ojarjur:
Thanks for reporting the issue.

Digging in to the source code, it seems that different implementations of the `retriever` method make different assumptions about the meaning of the second argument. `google.datalab.bigquery.Table` is the only one that assumes it is a running count of items. `google.datalab.bigquery.Datasets`, `google.datalab.ml.Models`, `google.datalab.ml.ModelVersions`, and `google.datalab.ml.Jobs` all assume it is a page size (which is the current behavior). The other clients in the `pydatalab` library all ignore the second argument.

As it is, simply changing the second argument to implement a running count would not be sufficient. We need to first make all of the retriever methods consistent with the meaning of the second argument.

I'll put together a PR to update the retriever methods.

j-planet:
Thanx @ojarjur . Making all of the retriever methods consistent with the meaning of the second argument is indeed a much better fix! Being a long-time fan of GCP products, I'm surprised that bugs of this level of severity exists in paid products as popular as `pydatalab`.

PR #687, opened 2018-05-28
Bug fix: google.datalab.utils.Iterator.__iter__ stuck in "infinite" loop

j-planet:
details here https://github.com/googledatalab/pydatalab/issues/686


j-planet:
`I signed it!`

PR #689, opened 2018-06-02
Fix an MLToolbox preprocessing failure caused by pickling of TF objects.

qimingj:
It seems happening after we upgrade DataFlow version to 2.0. The issue happens when:

1. You have "import tensorflow as tf" in a DoFn's __init__.
2. At run time, DataFlow initializes the DoFn first, and then pickle and unpickle the instance to multiple workers. That means it needs to pickle the state including imports.
3. Under tensorflow there are some objects that are not pickle friendly. So pickling everything under tensorflow namespace causes it to fail at run time.

The issue also occur if you have global "import tensorflow" statement.

The fix is to move tensorflow import into either DoFn's process(), or start_bundle() function. That way the import occurs for each instance on each worker and no tensorflow pickling is needed.

### DIFF
--- a/solutionbox/image_classification/mltoolbox/image/classification/_preprocess.py
+++ b/solutionbox/image_classification/mltoolbox/image/classification/_preprocess.py
@@ -24,13 +24,10 @@ import cStringIO
 import logging
 import os
 from PIL import Image
-import tensorflow as tf
 
 from . import _inceptionlib
 from . import _util
 
-
-slim = tf.contrib.slim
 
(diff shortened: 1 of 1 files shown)


ojarjur:
This looks good, but can you add some details about the root cause of the issue?

I'd like to have the lessons from your investigation documented in the code history.

qimingj:
Added in the code change comments (or you think better in code inline comments?). Thanks for reminding.

PR #709, opened 2018-09-25
utils: fix error handling for http requests in utils in py3

alienczf:
Minor bug in http request error parsing happens as `bytes` is `basestring` type in py3 and does not want to be split by an `'\n'` string. Introducing a simple fix for this by forcing the case (inelegant solution as opposed to actually decoding the message. Though that will take a bit more more type detection code)

Updated test to be both 2 and 3 compatible and reflects this error for py3 accurately. On a side note, I believe most of the payload comes out of py3 builtins as `bytes` so it's quite fortunate that we have to ran into any issues right now. We should update tests to reflect this in order to ensure that the api is py3 compatible too

```
...
  File "/root/.local/lib/python3.5/site-packages/google/datalab/utils/_http.py", line 158, in request
    raise RequestException(response.status, content)
  File "/root/.local/lib/python3.5/site-packages/google/datalab/utils/_http.py", line 55, in __init__
    lines = content.split('\n') if isinstance(content, basestring) else []
TypeError: a bytes-like object is required, not 'str'"
```

### DIFF
--- a/google/datalab/utils/_http.py
+++ b/google/datalab/utils/_http.py
@@ -52,7 +52,7 @@ class RequestException(Exception):
         error = error['errors'][0]
       self.message += ': ' + error['message']
     except Exception:
-      lines = str(content).split('\n') if isinstance(content, basestring) else []
+      lines = content.decode('utf-8').split('\n') if isinstance(content, basestring) else []
       if lines:
         self.message += ': ' + lines[0]
 
--- a/google/datalab/utils/_http.py
+++ b/google/datalab/utils/_http.py
@@ -44,15 +44,17 @@ class RequestException(Exception):
     self.message = 'HTTP request failed'
     # Try extract a message from the body; swallow possible resulting ValueErrors and KeyErrors.
     try:
-      error = json.loads(content, encoding='UTF-8')['error']
+      if isinstance(content, str):
+        error = json.loads(content)['error']
+      else:
+        error = json.loads(str(content, encoding='UTF-8'))['error']
       if 'errors' in error:
         error = error['errors'][0]
-      self.message += ': ' + error['message']
+      self.message += ': {}'.format(error['message'])
     except Exception:
       lines = content.splitlines() if isinstance(content, basestring) else []
       if lines:
-        msg = lines[0]
-        self.message += ': ' + (msg.decode('UTF-8') if isinstance(msg, bytes) else msg)
+        self.message += ': {}'.format(lines[0])
 
   def __str__(self):
     return self.message
--- a/google/datalab/utils/_http.py
+++ b/google/datalab/utils/_http.py
@@ -52,7 +52,7 @@ class RequestException(Exception):
         error = error['errors'][0]
       self.message += ': ' + error['message']
     except Exception:
-      lines = content.split('\n') if isinstance(content, basestring) else []
+      lines = str(content).split('\n') if isinstance(content, basestring) else []
       if lines:
         self.message += ': ' + lines[0]
 
--- a/google/datalab/utils/_http.py
+++ b/google/datalab/utils/_http.py
@@ -44,17 +44,15 @@ class RequestException(Exception):
     self.message = 'HTTP request failed'
     # Try extract a message from the body; swallow possible resulting ValueErrors and KeyErrors.
     try:
-      if type(content) == str:
-        error = json.loads(content)['error']
-      else:
-        error = json.loads(str(content, encoding='UTF-8'))['error']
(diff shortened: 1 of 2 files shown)


alienczf:
@yebrahim let me know if I should open up an issue for documentation or is the information in this MR sufficient

google/datalab/utils/_http.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, status, content):
         error = error['errors'][0]
       self.message += ': ' + error['message']
     except Exception:
-      lines = content.split('\n') if isinstance(content, basestring) else []
+      lines = str(content).split('\n') if isinstance(content, basestring) else []
```
yebrahim: This doesn't seem like the right way to handle unicode string, I think it must be decoded, otherwise the string isn't really split:
```
>>> print(str(b'hello\nworld').split('\n')) .            # ["b'hello\\nworld'"]
>>> print(b'hello\nworld'.decode('utf-8').split('\n'))   # ['hello', 'world']
```

alienczf: sorry about the slow response.. yes you're right of course.. wasn't sure that I can assume the utf8 encoding so a string cast is the most "robust" implementation (i.e. not crashing the code when encountering bytes outside the charset but can end up with garbled messages). If we can assume utf8 yes you're right. I've updated the code and pushed the change

google/datalab/utils/_http.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, status, content):
         error = error['errors'][0]
       self.message += ': ' + error['message']
     except Exception:
-      lines = content.split('\n') if isinstance(content, basestring) else []
+      lines = content.decode('utf-8').split('\n') if isinstance(content, basestring) else []
```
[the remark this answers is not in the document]
alienczf: I think the content arrives in python3 as bytes (bytestring) and not as string. That was the original error given that `b'bytes'.split('\n')` is invalid in python3.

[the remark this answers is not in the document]
yebrahim: I don't know if this is a good assumption to make here. It's probably better to make this resilient to both types of content. This will also make the testing situation a lot easier.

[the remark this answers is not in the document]
alienczf: Right.. Let me have a think and do a bit of digging into python core and see if there's a better way to handle this..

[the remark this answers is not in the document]
yebrahim: Thanks for working on this @alienczf!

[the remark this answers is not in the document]
alienczf: added a small workaround to split first and then cast.. it's still not watertight in terms of handling 'python2 bytes'

google/datalab/utils/_http.py
```diff
@@ -44,17 +44,15 @@ def __init__(self, status, content):
     self.message = 'HTTP request failed'
     # Try extract a message from the body; swallow possible resulting ValueErrors and KeyErrors.
     try:
-      if type(content) == str:
-        error = json.loads(content)['error']
-      else:
-        error = json.loads(str(content, encoding='UTF-8'))['error']
+      error = json.loads(content, encoding='UTF-8')['error']
```
yebrahim: Why make this change? It seems the old if/else is there for py2,3 compatibility.

google/datalab/utils/_http.py
```diff
@@ -44,17 +44,15 @@ def __init__(self, status, content):
...
       if 'errors' in error:
         error = error['errors'][0]
       self.message += ': ' + error['message']
     except Exception:
-      lines = content.split('\n') if isinstance(content, basestring) else []
+      lines = content.splitlines() if isinstance(content, basestring) else []
```
yebrahim: Looks like this is the only needed fix here, why are the lines below needed?

google/datalab/utils/_http.py
alienczf: The old code did not attempt to decode the content in python3 and the change was necessary. However, the json loads api handles all cases of string, bytes and basestring and correctly ignores the encoding flag in python3 if it doesn't need it.

I can revert the change or add new test cases to make sure that this implementation works for relevant python versions we need to support.

google/datalab/utils/_http.py
alienczf: Can't concat (+) bytes to str if content is bytes in python3.

yebrahim:
Merged. It would be great if you can fix the PR description to reflect the latest changes, just to make it easier for anyone reviewing the change later.
Thanks!

PR #702, opened 2018-08-14
Upgrade google.cloud.monitoring

rileyjbauer:
See relevant `google-cloud-python` repo directory:
https://github.com/GoogleCloudPlatform/google-cloud-python/tree/master/monitoring/google/cloud

This should address the build failures that were blocking https://github.com/googledatalab/pydatalab/pull/700

PR #733, opened 2020-07-07
chore(deps): widen pin for google-cloud-monitoring

busunkim96:
This pin is overly restrictive and will likely lead to dependency conflicts with other libraries in the near future.


busunkim96:
Hmm judging from the contents of #702 it looks like this requires additional code changes.

PR #755, opened 2021-12-20
Add cloud ml dependency

eyadsibai:


setup.py
```diff
@@ -83,6 +83,9 @@
     'scikit-learn==0.17.1',
     'ipykernel==4.4.1',
   ],
+  dependency_links=[
```
eyadsibai, line 86: this should be removed

